PR #17, opened 2020-08-24
Use correct function for creating arena

### LINKED ISSUE #16: MMQA test function/7.c fails with RESOURCE
```
$ perl test/qa run function/7.c
% ERROR
!error=true
!errtext=create: RESOURCE
```
The problem is here in test/function/7.c:
```
die(mps_arena_create(&arena, mps_arena_class_vm(), mps_args_none), "create");
```
This is using the (deprecated) varargs interface [`mps_arena_create()`](https://www.ravenbrook.com/project/mps/master/manual/html/topic/deprecated.html#c.mps_arena_create) but passing `mps_args_none` for the `arena_size` parameter. This causes the address of `mps_args_none` to be interpreted as the requested size of the virtual memory arena.

Is there a way we could have caught this mistake earlier? I can't think of one: any value is acceptable as the `arena_size` parameter for a virtual memory arena -- we round the value up to a multiple of the page size, following the general principle that sizes need to be automatically rounded to acceptable value in order to make it easy to write portable programs. So it is not clear what we could do here to detect a bad value. Possibly we could check specifically for `mps_args_none`? But it's not impossible for the linker to put that at an address that equals a legitimate size.

The most effective thing we could do to avoid this mistake is to remove `mps_arena_create()` from the API.

gareth-rees:
Fixes #16 (MMQA test function/7.c fails with RESOURCE)

Since we are passing a list of keyword arguments, we must call the keyword-taking function (`mps_arena_create_k`) and not the deprecated varargs function (`mps_arena_create`).

### DIFF
--- a/test/function/7.c
+++ b/test/function/7.c
@@ -19,7 +19,7 @@ static void test(void *stack_pointer)
 
  for (p=0; p<ARENAS; p++)
  {
-  die(mps_arena_create(&arena, mps_arena_class_vm(), mps_args_none), "create");
+  die(mps_arena_create_k(&arena, mps_arena_class_vm(), mps_args_none), "create");
   if (p > 0)
    mps_arena_destroy(arena1);
   arena1=arena;


UNAA008 (approved):

UNAA008 (commented):
mps_arena_create was deprecated in 1.112, so we might consider removing it in the next release.
GDR comments that only long-term users of the MPS might still have it in their code so we might ask those that we know before making this decision.

PR #18, opened 2020-08-24
Set SA_RESTART flag when calling sigaction().

### LINKED ISSUE #9: Signal handling results in mutator getting EINTR for system calls
On Posix operating systems (FreeBSD and Linux), the MPS handles barrier hits by installing a signal handler for `SIGSEGV` (see `ProtSetup()` in protsgix.c), and suspends and resumes threads by calling `pthread_kill()` to send a signal to the thread (see `PThreadextSuspend()` and `PThreadextResume()` in pthrdext.c) which is caught by the signal handler installed on the thread (see `sigHandle()` in protsgix.c).

If any of these signals is received while the mutator was in any of the following system calls:

> `open`, `read`, `write`, `sendto`, `recvfrom`, `sendmsg`, `recvmsg`, `wait`, `ioctl`

then the system call may fail with `EINTR`. Morally speaking, it is a bug in the mutator if it fails to deal with this condition, since programs coding against the Posix API need to be prepared to handle `EINTR` from these system calls, but inevitably many don't. The MPS could avoid breaking these programs by setting the `SA_RESTART` flag when calling `sigaction()`. There are three places where this is necessary:

1. `ProtSetup()` in protsgix.c
2. `PThreadextModuleInit()` in pthrdext.c (two `sigaction()` calls here)

There are two calls to `sigaction()` in `sigHandle()` in protsgix.c, but it is not necessary to set the `SA_RESTART` flag because we are re-using the saved `struct sigaction` from the mutator and in that case we must use the flags that the mutator wanted.

(Issue originally identified by Filip Strömbäck.)

gareth-rees:
Fixes #9 (Signal handling results in mutator getting `EINTR` for system calls)

Ensure that if a mutator thread is blocked in a system call when the MPS handles a signal, the system call will not fail with `EINTR` but instead will be restarted.

Add a test case for the thread suspend and resume signals.

### DIFF
--- a/design/thread-manager.txt
+++ b/design/thread-manager.txt
@@ -63,6 +63,13 @@ request.dylan.160022_ and request.mps.160093_.)
 .. _request.dylan.160022: https://info.ravenbrook.com/project/mps/import/2001-11-05/mmprevol/request/dylan/160022
 .. _request.mps.160093: https://info.ravenbrook.com/project/mps/import/2001-11-05/mmprevol/request/mps/160093/
 
+_`.req.thread.intr`: It would be nice if on POSIX systems the MPS does
+not cause system calls in the mutator to fail with EINTR due to the
+MPS thread-management signals being delivered while the mutator is
+blocked in a system call. (See `GitHub issue #9`_.)
+
+.. _GitHub issue #9: https://github.com/ravenbrook/mps/issues/9
+
 
 Design
 ------
@@ -94,6 +101,25 @@ _`.sol.thread.term.attempt`: Nonetheless, the thread manager makes a
 "best effort" to continue running after detecting a terminated thread,
 by moving the thread to a ring of dead threads, and avoiding scanning
 it. This might allow a malfunctioning client program to limp along.
+
+_`.sol.thread.intr`: The POSIX specification for sigaction_ says that
+if the ``SA_RESTART`` flag is set, and if "a function specified as
+interruptible is interrupted by this signal, the function shall
+restart and shall not fail with ``EINTR`` unless otherwise specified."
+
+.. |sigaction| replace:: ``sigaction()``
+.. _sigaction: https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigaction.html
+
+_`.sol.thread.intr.linux`: Linux does not fully implement the POSIX
+specification, so that some system calls are "never restarted after
+being interrupted by a signal handler, regardless of the use of
+SA_RESTART; they always fail with the error EINTR when interrupted by
+a signal handler". The exceptional calls are listed in the |signal|_
+manual. There is nothing that the MPS can do about this except to warn
+users in the reference manual.
+
+.. |signal| replace:: signal(7)
+.. _signal: https://man7.org/linux/man-pages/man7/signal.7.html
 
 
 Interface
--- a/manual/source/release.rst
+++ b/manual/source/release.rst
@@ -57,6 +57,15 @@ Interface changes
    .. _struct: https://docs.python.org/3/library/struct.html#struct.unpack
 
 
+Other changes
+.............
+
+#. On FreeBSD and Linux, if the MPS handles a signal while the client
+   program is blocked in a system call, the system call is
+   automatically restarted and does not fail with ``EINTR``. See
+   :ref:`topic-thread-signal`.
+
+
 .. _release-notes-1.117:
 
 Release 1.117.0
--- a/test/function/236.c
+++ b/test/function/236.c
@@ -1,0 +1,155 @@
+/*
+TEST_HEADER
+ id = $Id$
+ summary = regression test for GitHub issue #9
+ language = c
+ link = testlib.o myfmt.o
+ parameters =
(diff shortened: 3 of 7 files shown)


test/function/236.c
```diff
@@ -0,0 +1,130 @@
...
+  die(mps_arena_create_k(&arena, mps_arena_class_vm(), mps_args_none), "mps_arena_create_k");
+  die(mps_fmt_create_A(&format, arena, &fmtA), "create format");
+  MPS_ARGS_BEGIN(args) {
+     MPS_ARGS_ADD(args, MPS_KEY_FORMAT, format);
+     die(mps_pool_create_k(&pool, arena, mps_class_amc(), args), "mps_pool_create_k");
+  } MPS_ARGS_END(args);
+
+  /* Create data structures for synchronizing with the thread. */
+  e = pipe(pipefd);
+  asserts(e == 0, "pthread_create: %s", strerror(errno));
```
gareth-rees, line 90: Typo: "pthread_create" for "pipe" in the assertion text

test/function/236.c
```diff
@@ -0,0 +1,130 @@
...
+{
+  test_data_t *data = arg;
+  mps_thr_t thread;
+  int e;
+  char c;
+
+  die(mps_thread_reg(&thread, data->arena), "mps_thread_reg");
+  e = sem_post(data->sem);
+  asserts(e == 0, "sem_post: %s", strerror(errno));
+  e = read(data->fd, &c, 1);
```
gareth-rees, line 64: Add a comment mentioning that we know there's a race here.

gareth-rees:
@UNAA008 I couldn't figure out a way to make a system call fail with `EINTR` due to a barrier hit -- I tried `read` and `write` with the buffer in protected memory, but these fail with `EFAULT` when the barrier is hit, and they are not restarted. So I'm proposing this be merged as-is.

design/thread-manager.txt
```diff
@@ -95,6 +102,14 @@ _`.sol.thread.term.attempt`: Nonetheless, the thread manager makes a
...
 it. This might allow a malfunctioning client program to limp along.
 
+_`.sol.thread.intr`: The POSIX specification for sigaction_ says that
+if the ``SA_RESTART`` flag is set, and if "a function specified as
+interruptible is interrupted by this signal, the function shall
+restart and shall not fail with ``EINTR`` unless otherwise specified."
+
+.. |sigaction| replace:: ``sigaction()``
+.. _sigaction: https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigaction.html
+
```
fstromback, line 112: It is worth noting here (or somewhere else as well perhaps) that certain system calls do not respect the `SA_RESTART` flag. On Linux, these mostly involve timekeeping, for example `select`, `usleep`, `epoll_wait`, and pthread functions like `sem_timedwait`. This is described in `man 7 signal` on my machine, also linked here (I could not find it in pubs.opengroup.org): [man7.org](https://man7.org/linux/man-pages/man7/signal.7.html)

test/function/236.c
```diff
@@ -0,0 +1,144 @@
...
+  data.fd = pipefd[0];
+
+  /* Start the thread. */
+  e = pthread_create(&pthread, NULL, thread_fn, &data);
+  asserts(e == 0, "pthread_create: %s", strerror(e));
+
+  /* Wait for the thread to register itself. */
+  e = sem_wait(&sem);
+  asserts(e == 0, "sem_wait: %s", strerror(errno));
+
```
fstromback, line 118: In some similar tests I have, I use a sleep of a few milliseconds here, to make it more likely that the other thread is being blocked in the `read` system call before continuing in this thread. Thus, making it "more likely" to trigger the desired problem. Sadly, it adds some time to the test cases, and does not give any strict guarantees regarding correctness.

gareth-rees: In practice I found that `mps_arena_collect()` has enough work to do that the `read` system call had always been entered by the time the MPS suspends threads. But it does no harm to add a millisecond of sleep here, so I did that.

design/thread-manager.txt
gareth-rees: I added some discussion of this issue to the "Signal and exception handling issues" section of the reference manual.

test/function/236.c
gareth-rees: In some other work (not MPS-related) I've used the "poll /proc/PID/status until the process enters the required state" technique for synchronization in difficult cases, but that's on a project which is closely tied to Linux so that there's no portability concern.

manual/source/release.rst
```diff
@@ -4,6 +4,22 @@ Release notes
...
+Release 1.119.0
+---------------
+
+Other changes
+.............
+
+#. On FreeBSD and Linux, if the MPS handles a signal while the client
+   program is blocked in a system call, the system call is
+   automatically restarted and does not fail with ``EINTR``. See
+   `GitHub issue #9`_.
```
gareth-rees, line 18: A better link would be to "Signal and exception handling issues" in the reference manual.

PR #32, opened 2020-08-30
Save and restore errno in signal handlers.

### LINKED ISSUE #10: MPS signal handlers might modify errno
On Posix operating systems (FreeBSD and Linux), the MPS handles barrier hits, and suspends and resumes threads, by installing signal handlers with `sigaction()`. The signal handlers may call libc functions like `mprotect()`, `mmap()` and so on, that set `errno`.

The [Posix specification for `sigaction()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigaction.html) says:

> Note in particular that even the "safe" functions may modify `errno`; the signal-catching function, if not executing as an independent thread, should save and restore its value in order to avoid the possibility that delivery of a signal in between an error return from a function that sets `errno` and the subsequent examination of `errno` could result in the signal-catching function changing the value of `errno`.

This applies to the following signal handlers:

1. `sigHandle()` in protsgix.c
2. `suspendSignalHandler()` in pthrdext.c
3. `resumeSignalHandler()` in pthrdext.c (this can in theory call `mps_lib_assert_fail()` if the assertion fails, and in the hot and rash varieties the application can continue from the assertion failure)

gareth-rees:
Fixes #10 (MPS signal handlers might modify errno)

### DIFF
--- a/code/misc.h
+++ b/code/misc.h
@@ -235,6 +235,16 @@ typedef const struct SrcIdStruct {
 #define BS_BITFIELD(ty, s)      BITFIELD(ty ## Set, (s), ty ## LIMIT)
 
 
+/* Save and restore errno. See <design/thread-manager#.sol.thread.errno>
+ *
+ * These macros must be used after #include <errno.h> so that errno is
+ * defined.
+ */
+
+#define ERRNO_SAVE BEGIN int _saved_errno = errno; BEGIN
+#define ERRNO_RESTORE END; errno = _saved_errno; END
+
+
 #endif /* misc_h */
 
 
--- a/code/protsgix.c
+++ b/code/protsgix.c
@@ -28,6 +28,7 @@
 
 #include "prmcix.h"
 
+#include <errno.h>     /* for errno */
 #include <signal.h>    /* for many functions */
 #include <ucontext.h>  /* for ucontext_t */
 #include <unistd.h>    /* for getpid */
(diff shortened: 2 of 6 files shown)


code/protsgix.c
```diff
@@ -71,47 +72,55 @@ static struct sigaction sigNext;
...
+    /* get an accurate context (the MPS would fail if it were the second in */
+    /* line) but it's the best we can do. */
+
+    e = sigaction(PROT_SIGNAL, &sigNext, &sa);
+    AVER(e == 0);
+    e = sigemptyset(&asigset);
+    AVER(e == 0);
+    e = sigaddset(&asigset, PROT_SIGNAL);
+    AVER(e == 0);
+    e = sigprocmask(SIG_UNBLOCK, &asigset, &oldset);
```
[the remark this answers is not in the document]
gareth-rees: It can be `sigprocmask` not `pthread_sigmask` because this code runs even in the single-threaded case and because we don't care what thread the SEGV is delivered to.

code/misc.h
```diff
@@ -235,6 +235,12 @@ typedef const struct SrcIdStruct {
 #define BS_BITFIELD(ty, s)      BITFIELD(ty ## Set, (s), ty ## LIMIT)
 
 
+/* Save and restore errno. See <design/thread-manager#.sol.thread.errno> */
+
+#define ERRNO_SAVE BEGIN int _mps_saved_errno = (errno); BEGIN
+#define ERRNO_RESTORE END; (errno) = _mps_saved_errno; END
+
```
UNAA008, line 242: I had fun researching the reasons for the BEGIN/END constructs.
I'm a bit puzzled by parantheses around errno.  Is that defensive against whatever macro errno might be in some implementations?

gareth-rees: You're right, the parentheses are unnecessary: this macro can only be used after `#include <errno.h>` so `errno` should be defined correctly by the header. I'll add a note and remove the parentheses.

UNAA008 (approved):
FTR, GDR walked me through these changes on Sunday and I've subsequently worked my way through them since.  I am not a C or MPS expert but the changes look sound to me and I'm approving on that basis.

PR #34, opened 2020-08-31
New API function mps_pool_walk

### LINKED ISSUE #33: Walk interface is not suitable for common applications
The MPS offers an object-walking interface via [`mps_arena_formatted_objects_walk()`](https://www.ravenbrook.com/project/mps/master/manual/html/topic/format.html#c.mps_arena_formatted_objects_walk). But this is unsuitable for certain applications:

1. You can't update references in the object being walked (because there might be a summary on the segment and the summary will need to be updated if the references change). This affects applications like hot reloading of code, where the run-time might need to write all objects belonging to a class (say) so that they point to the new class definitions.

2. The callback is called for every object, so that there is per-object function call overhead.

3. It requires a Walk method for every segment class, that needs to be very similar to the existing Scan and Reclaim methods, but which can't share any code with them.

Here are some sketch requirements and design for a new function solving all these problems:

### Requirements

req.walk.all: It must be possible for the client program to visit all automatically managed formatted objects using a callback.

req.walk.assume-format: The callback should not need to switch on the format, as this may be awkward in a program which has modules using different pools with different formats.

req.walk.examine: It must be possible for the callback to examine other automatically managed memory while walking the objects.

req.walk.modify: It must be possible for the callback to modify the references in the objects.

req.walk.overhead: The overhead of calling the callback should be minimized.

req.walk.perf: The performance of subsequent collections should not be affected.

req.walk.closure: The callback must have access to arbitrary data from the caller.

req.walk.maint: The interface should be easy to implement and maintain.

### Design

A new public function `mps_pool_walk()` visits the live formatted objects in an automatically managed pool.

sol.walk.all: The client program must know which pools it has created so it can call mps_pool_walk for each pool.

sol.walk.assume-format: All objects in a pool share the same format, so the callback does not need to switch on the format.

sol.walk.examine: `mps_pool_walk()` must only be called when the arena is parked, and so there is no read barrier on any object.

sol.walk.modify: `mps_pool_walk()` arranges for write-protection to be removed from each segment while it is being walked and restored afterwards if necessary.

sol.walk.overhead: The callback is called for contiguous regions of formatted objects (not just for each object) where possible so that the per-object function call overhead is minimized.

sol.walk.perf: The callback uses the scanning protocol so that every reference is fixed and the summary is maintained.

sol.walk.closure: `mps_pool_walk()` takes a closure pointer which is passed to the callback.

sol.walk.maint: This design reuses the scanning protocol so we can provide a generic implementation that iterates over the ring of segments in the pool. We set up an empty white set in the `ScanState` so that the `MPS_FIX1` test always fails and `_mps_fix2` is never called. We can refactor the way that pools call the scan method so that we can pass in a scan method via the scan state. (This would be an equivalent of `FormatScan()` in the usual scanning case.) There is a small difficulty in the case of LO in that it has formatted objects but its segments don’t have a scan method but we can afford to write one.

### Interface

`mps_res_t mps_pool_walk(mps_pool_t pool, mps_area_scan_t scan_area, void *closure);`

Visit all regions of formatted objects in a pool. The pool must be automatically managed. The pool’s arena must be parked.

`pool` is the pool whose formatted objects you want to visit.

`scan_area` is an area scanning function. See [Area scanners](https://www.ravenbrook.com/project/mps/master/manual/html/topic/scanning.html#c.mps_area_scan_t).

`closure` is an arbitrary pointer that will be passed to `scan_area`.

gareth-rees:
Fixes #33 (Walk interface is not suitable for common applications)


code/trace.c
```diff
@@ -114,6 +135,42 @@ void ScanStateInit(ScanState ss, TraceSet ts, Arena arena,
...
+{
+  Format format;
+  AVERT(Seg, seg);
+
+  ScanStateInit(ss, ts, arena, rank, white);
+  if (PoolFormat(&format, SegPool(seg))) {
+    ss->formatScan = traceFormatScan;
+    ss->formatScanClosure = format;
+  } else {
+    ss->formatScan = traceNoScan;
```
gareth-rees, line 169: This is not necessary (already set by `ScanStateInit`)

code/trace.c
```diff
@@ -1561,7 +1530,15 @@ Res TraceScanFormat(ScanState ss, Addr base, Addr limit)
    * ss->formatScan. */
   ss->scannedSize += AddrOffset(base, limit);
 
-  return ss->formatScan(&ss->ss_s, base, limit, ss->formatScanClosure);
+  switch (ss->callbackType)
+  {
+      default:
+          NOTREACHED;
```
gareth-rees, line 1536: Add a `return ResFAIL` here so that we can in theory continue from the assertion.

gareth-rees:
This pull request just passed its first anniversary. Many happy returns to #34!

fstromback:
Please tell me if there is anything I can do to help.

code/amcss.c
```diff
@@ -128,6 +128,24 @@ static void test_stepper(mps_addr_t object, mps_fmt_t fmt, mps_pool_t pool,
...
 
 
+/* area_scan -- area scanning function for mps_pool_walk */
+
+static mps_res_t area_scan(mps_ss_t ss, void *base, void *limit, void *closure)
+{
+  unsigned long *count = closure;
+  mps_res_t res;
+  while (base < limit) {
+    mps_addr_t prev = base;
```
thejayps, line 138: Here and elsewhere I see void * and mps_addr_t used interchangeably, is there a reason to use void* when we have mps_addr_t? This assumes mps_addr_t will never change its underlying type (which I guess is a fairly safe assumption)

gareth-rees: The two types are used to make a distinction: `mps_addr_t` represents an address in memory that is managed by the MPS ([see the manual](https://www.ravenbrook.com/project/mps/master/manual/html/topic/interface.html#c.mps_addr_t)), while `void *` represents any address, that might or might not be in memory managed by the MPS.

In the case of amcss.c, `area_scan()` has to belong to the type `mps_area_scan_t` (so that its arguments are `void *`). Area scanning functions were introduced by @rptb1 to fix [job003668](https://www.ravenbrook.com/project/mps/issue/job003668/), but I am not sure why he chose to use `void *` for the area scanning arguments and not `mps_addr_t`.

@rptb1 Do you remember why you made this choice?

gareth-rees:
> Please tell me if there is anything I can do to help.

Since this pull request is (partly) aimed at solving your problem, maybe you can have a look and check that it would have done so! Also, if you'd like to review the changes, that would be helpful — in particular, is the documentation for the new feature clear and easy to follow?

fstromback:
Thank you for the updates!

I have looked through the suggested changes (mostly focusing on documentation), and it looks good to me. The documentation is clear and easy to follow. The only thing that might be worth pointing out in the documentation is that if the area scanning function modifies any references, it should scan them after they have been modified. If I understand the semantics of the summaries correctly, it would not break the MPS if the area scanning function scanned some references before *and* after they are updated, just cause some additional scanning in future collections.

One detail that might be relevant to document is whether or not it is possible to allocate memory from the MPS from the area scanning function. I assume the answer is "no", since the MPS is not reentrant in general. If that is the case, the documentation is fine as it is. If it is possible in some circumstances, it would be useful to note that there (if this would be possible, I could probably avoid an extra heap traversal in some cases. I assume other use-cases would benefit as well).

I think this PR is enough to implement what I need in terms of updating the heap. With it, I will be able to find and update all references in the entire program, so I think it would cover almost any use case I might come up with as well. Allocating memory in the area scan function would probably help in some situations, but I am able to work around that limitation, so this PR does not need to be extended for that.

gareth-rees:
> The only thing that might be worth pointing out in the documentation is that if the area scanning function modifies any references, it should scan them after they have been modified.

Good point — I will emphasize that in the docs.

> One detail that might be relevant to document is whether or not it is possible to allocate memory from the MPS from the area scanning function. I assume the answer is "no", since the MPS is not reentrant in general. If that is the case, the documentation is fine as it is. If it is possible in some circumstances, it would be useful to note that there (if this would be possible, I could probably avoid an extra heap traversal in some cases. I assume other use-cases would benefit as well).

That's right — it's not possible to call back into the MPS from the scanning function. This is mentioned in the [Cautions](https://www.ravenbrook.com/project/mps/master/manual/html/topic/format.html#cautions) section of the Object formats chapter, but there ought to be something similar in the Scanning chapter since the constraints are not identical.

> Allocating memory in the area scan function would probably help in some situations

This is interesting — can you describe the use case? (For future reference.)

fstromback:
> >  Allocating memory in the area scan function would probably help in some situations
>
> This is interesting — can you describe the use case? (For future reference.)

This would be relevant when a field has been added to a type, so that I need to grow objects of that type. Currently, I can solve this by: 1: walking the heap to find all instances of the object and recording them somewhere. 2: allocating copies of these objects (after walking the heap). 3: walking the heap again, updating all references to the copied objects. 

If it was possible to allocate memory while walking the heap, I could probably merge it into one heap walk. Each time "fix" would be called, I examine the object referred by the pointer, if it is an object I need to update, I make a larger copy with the required changes and replace the original with a forwarding object. When I encounter a pointer to a forwarding object, I can then replace the pointers for remaining instances quite easily (this process is quite similar to what I understand happens in the amc pool during a GC). Note: it might not be possible to use forwarding objects in this manner, as it might confuse the MPS later on, but I can make something that does not look like a forwarding object to the MPS, or use a separate data structure for the same thing.

There are still some details that I need to work out here to get all edge cases correct, for example when I have a linked structure of objects that all need to be copied. The details for this depends on whether or not allocations done during the heap walk will be walked or not. Regardless of the semantics, I think it would be doable.

But don't worry too much about this now. The three step process is probably fast enough (updating a running program does not happen too often, so if it takes an extra 100 ms it is fine), and I can probably (ab)use the allocation point protocol to get a buffer I can allocate some objects from without reentering the MPS as a starting point for further optimization/experimentation if I need more speed.

rptb1:
As part of work on #110 Ravenbrook is assessing the risk to Configura of all changes since 1.115 .  This is a risky change because it touches the tracer.  It's an important change because it could replace the custom "transforms" code deployed at Configura as part of resolving #110.

Therefore we'd like to put this through formal review, even though it's already merged.

Executing [proc.review.entry](https://github.com/Ravenbrook/mps/blob/936f0585c61031e3b19436fd889721f6ab048501/procedure/review.rst#51-review-entry).

1. Start time 18:37.
2. proc.review.entry.criteria: I'm applying [entry.universal](https://github.com/Ravenbrook/mps/blob/936f0585c61031e3b19436fd889721f6ab048501/procedure/entry.universal.rst), [entry.design](https://github.com/Ravenbrook/mps/blob/936f0585c61031e3b19436fd889721f6ab048501/procedure/entry.design.rst), [entry.impl](https://github.com/Ravenbrook/mps/blob/936f0585c61031e3b19436fd889721f6ab048501/procedure/entry.impl.rst).
3. entry.universal.author: I've asked @gareth-rees to agree to the review.  I'm going to assume agreement for now since he previously requested review.on 2020-09-01.
4. entry.universal.source-available: Source documents include a detailed mail discussion (see below) and the Configura custom transforms [design](https://info.ravenbrook.com/project/mps/custom/cet/main/design/transform.txt) and implementation.  All available.  Reading the mail thread will be part of [proc.review.plan.homework](https://github.com/Ravenbrook/mps/blob/936f0585c61031e3b19436fd889721f6ab048501/procedure/review.rst#plan-homework).
5. entry.universal.source-reviewed: Transforms were reviewed and have been in deployment for years without any failures reported.  Other sources are unreviewed.
6. entry.design.rfc: This branch introduces [design.mps.walk](https://github.com/Ravenbrook/mps/pull/34/files#diff-3b61a090dd9c337528d090be7917b61b6eca75db35a301c9d5e2f05474c73d7d) and changes other designs.  I believe these have had informal review, partly from @fstromback , and I'll include them in [proc.review.plan.homework](https://github.com/Ravenbrook/mps/blob/936f0585c61031e3b19436fd889721f6ab048501/procedure/review.rst#plan-homework).
7. Entry passed.
8. End time 18:52.
9. Entry took 15 minutes.

The mail thread leading to this work, discovered by `p4 grep`, is:
- https://info.ravenbrook.com/mail/2020/08/20/21-01-34/0.txt
- https://info.ravenbrook.com/mail/2020/08/23/12-48-45/0.txt
- https://info.ravenbrook.com/mail/2020/08/23/13-46-40/0.txt
- https://info.ravenbrook.com/mail/2020/08/23/16-19-00/0.txt
- https://info.ravenbrook.com/mail/2020/08/25/07-06-48/0.txt
- https://info.ravenbrook.com/mail/2020/08/25/15-26-47/0.txt
- https://info.ravenbrook.com/mail/2020/08/26/16-43-11/0.txt
- https://info.ravenbrook.com/mail/2020/08/26/21-48-50/0.txt
- https://info.ravenbrook.com/mail/2020/08/30/15-01-23/0.txt
- https://info.ravenbrook.com/mail/2020/08/30/19-17-03/0.txt
- https://info.ravenbrook.com/mail/2020/08/31/07-05-28/0.txt
- https://info.ravenbrook.com/mail/2020/08/31/07-39-12/0.txt
- https://info.ravenbrook.com/mail/2020/08/31/16-31-32/0.txt
- https://info.ravenbrook.com/mail/2020/09/03/13-02-25/0.txt
- https://info.ravenbrook.com/mail/2020/09/03/19-49-47/0.txt
- https://info.ravenbrook.com/mail/2020/09/03/21-12-11/0.txt
- https://info.ravenbrook.com/mail/2020/09/04/07-46-39/0.txt

rptb1:
[job004090](https://info.ravenbrook.com/project/mps/issue/job004090/) is a source for the review, because we must ensure we don't regress.

thejayps:
Executing [proc.review.plan](https://github.com/Ravenbrook/mps/blob/4634a33f18dc6b30675235405ca43a3636f4942b/procedure/review.rst#52-review-planning):

1. Started 09:23.
2. Pairing this procedure with @thejayps so that he can observe how I do it.  This is quite a complicated change so there are things to discuss.  These notes will be more detailed and discursive than usual.
3. Eyeballing the diffs, we can see that this is high risk change (touches trace.c and pools).
4. Size of change is +1000 -300 ish.  Mostly code change.  Design doc change is just a copy of the issue plus references updated.  The manual has some things moved around and about 100 lines of new doc.  About 200 lines of new test case code.  About 100 lines of misc.  That leaves about 600 lines of code, so 1 h @ 10 loc/min.
5. Thinking about roles and tactics:
6. proc.review.role.check.correctness is most important.  Part of that will be to check against the older transforms code (which we know to work).
7. proc.review.role.check.consistency with documentation is very important.
8. proc.review.role.check.source with the original issues: there's @fstromback 's mail thread but most importantly, against Configura's requirements
9. That raises some extra proc.review.plan.homework: refreshing understanding Configura's requirements.  I'm pretty sure Configura would be OK with us documenting their requirements for this in public.  We should ask.
10. Thinking about skills and knowledge, roughly: @rptb1 for correctness, @thejayps for consistency, @UNAA008 on sources.  To be clear, we should all do the homework, but @UNAA008 's checking role will be to check between that stuff and the change.  Tracking down transforms and Configura requirements is probably homework for @rptb1 and @thejayps : check Configura current usage with them and source code .
11. proc.review.plan.tactics: Using 10 loc/minute rule: Code checking is about 1 h @ 10 loc/min for code correcness.  The rest is all smaller.  It's feasable to review this in one session.  We might spill over.
12. proc.review.plan.schedule: We discussed scheduling in a meeting yesterday.  Not this week.  Maybe Thursday 2023-03-02 13:00?
13. proc.review.plan.source: See https://github.com/Ravenbrook/mps/pull/34#issuecomment-1430226687 plus #33 plus documents @rptb1 digs up from Transforms during homework.
14. proc.review.plan.homework: We all need to read the requirements: mail thread, issue, whatever is implied by existing Configura usage. I don't think there's anything external to learn about.
15. @rptb1 will write an invitation mail later today.
16. Planning finished 10:01.
17. Planning took 30 mins.

thejayps:
> 4. Size of change is +1000 -300 ish.  Mostly code change.  Design doc change is just a copy of the issue plus references updated.  The manual has some things moved around and about 100 lines of new doc.  About 200 lines of new test case code.  About 100 lines of misc.  That leaves about 600 lines of code, so 1 h @ 10 loc/min.

Possible PI here to add more detail for size estimation.

thejayps:
More possible PI: Discussing how we paired on planning here, we could perhaps have a short guide on how to use MPS procedure documents effectively, so that PIs take hold.  Repeatedly re-reading the procedure and all of its bullet points is key -- treating the procedure as a checklist of what's been done more than set of steps.

UNAA008:
I read the mail thread with a focus on requirements.
Here's what I wrote
[Review_mps_pool_walk_2023-03.PDF](https://github.com/Ravenbrook/mps/files/10865310/Review_mps_pool_walk_2023-03.PDF)
Apologies that something has disabled the hyperlinks in this rendering.

UNAA008:
[RMPW_2023-03.ZIP](https://github.com/Ravenbrook/mps/files/10870294/RMPW_2023-03.ZIP)

GITHUB won't allow HTML attachments but this ZIP file contains the HTML and CSS.

rptb1:
Executing [proc.review.kickoff](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#53-review-kickoff)

1. Start time 15:03.
2. Logging meeting set for 16:15.
3. Kickoff took 9 mins.

code/trace.c
```diff
@@ -1473,16 +1516,36 @@ void TraceScanSingleRef(TraceSet ts, Rank rank, Arena arena,
 }
 
 
+/* TraceScanFormat -- scan a formatted area of memory for references
+ *
+ * This is a wrapper for format scanning functions, which should not
+ * otherwise be called directly from within the MPS.  This function
```
rptb1, line 1522: m. Why should they not be called? rule.generic.clear

code/walk.c
```diff
@@ -381,6 +381,123 @@ void mps_arena_roots_walk(mps_arena_t mps_arena, mps_roots_stepper_t f,
...
+  AVERT(Pool, pool);
+  AVER(FUNCHECK(area_scan));
+  /* closure is arbitrary and can't be checked */
+
+  AVER(ArenaGlobals(arena)->clamped);          /* .assume.parked */
+  AVER(arena->busyTraces == TraceSetEMPTY);    /* .assume.parked */
+
+  /* Synthesize a flipped trace with an empty white set. The empty
+   * white set means that the MPS_FIX1 test will always fail and
+   * _mps_fix2 will never be called. */
```
rptb1, line 439: m. Is this dependency documented at both ends? rule.code.deps.

manual/source/topic/pool.rst
```diff
@@ -170,3 +170,39 @@ Pool introspection
...
+
+    :c:data:`pool` is the pool whose formatted objects are visited.
+
+    :c:data:`scan_area` is an area scanning function. See
+    :ref:`topic-scanning-area`.
+
+    :c:data:`closure` is an arbitrary pointer that will be passed to
+    :c:data:`scan_area`.
+
+    The scanning function is called multiple times with disjoint areas
```
thejayps, line 190: m: clarity - although the documentation is clear in explaining what the scanner must do, it's less clear what the scanner might be expected or permitted to do in common applications.

UNAA008 (commented):
I spent most of the review time examining test code to see what requirements were explicitly checked.
AMCSSTH.C 
Checks that "walking works" while other threads continue to allocate.
I don't think walked objects are modified during walking.
Tests 104.C and 97.C are better documented and appear to address req.walk.all and req.walk.modify.

It isn't clear if req.walk.examine or req.walk.perf are tested.
I can't see any tests for serialisation.

In the documentation I noticed that 43.5 References contains a broken hyperlink.

thejayps (commented):
2 minor defects found: details are in in-line comments.

rptb1 (commented):
Executing [proc.review.check](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#54-review-checking)

1. C. In <https://info.ravenbrook.com/mail/2020/08/31/07-39-12/0/>, we should consider what it would take to walk objects without parking the arena.  It could be a test of multiple traces.
2. C. In <https://info.ravenbrook.com/mail/2020/08/23/16-19-00/0/>, we should study Storm's reloading and see if it could help Clasp or meet Clasp's loading requirements.
3. C. In <https://info.ravenbrook.com/mail/2020/08/25/07-06-48/0/> there is an implied requirement to remove all protection in the MPS. I think we provided a "portmortem" type of call to do this for Configura. The next collection will have to scan the world, but maybe that's acceptable sometimes.
4. m. design.mps.walk.sol.walk.all says it only visits live objects.  How does it determine that?  Does it mean reachable?  rule.generic.clear
5. m. <https://github.com/Ravenbrook/mps/blob/1dd501e7f23910f5ffac711bcdb43626b9e833eb/code/walk.c#L124> impl.c.walk.assume.parked does not explain *why* the root walker must be invoked with a parked arena.  rule.generic.clear
6. m. <https://github.com/Ravenbrook/mps/blob/1dd501e7f23910f5ffac711bcdb43626b9e833eb/code/walk.c#L124> implc.w.alk.assume.parked is assumed by the pool walker but only talks about the roots walker.  rule.generic.self
7. Checking incomplete but stopped at 16:13 for tea and logging.

design/walk.txt
```diff
@@ -0,0 +1,157 @@
...
+from the caller.
+
+_`.req.walk.maint`: The interface should be easy to implement and
+maintain.
+
+
+Design
+------
+
+A new public function ``mps_pool_walk()`` visits the live formatted
```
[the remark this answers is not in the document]
rptb1: Nm. This paragraph is not tagged.

rptb1:
Executing [proc.review.log](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#55-review-logging)

1. Start time 16:15.
2. Point of order. None of us have completed checking. @UNAA008 thinks about 80%, @thejayps thinks about 60%, @rptb1 thinks about 15% but of the most important code.  We'll log what we have, and schedule more time for review checking.
3. @UNAA008 would like to check in a pair to help with navigation.
4. Findings: @UNAA008 1 major, 1 minor, 2 questions. @thejayps 0 major, 2 minor. @rptb1 2 major, 11 minor, 4 comments.
5. @UNAA008 M. Not all requirements are tested. Not clear what rule that breaks.
6. @rptb1 Nm. Requirements are not linked to tests and tests are not documented in the design.
7. @rptb1 Nm. Test code is not linked to the requirements tested.
8. @rptb1 C. Pool walking is using the GC mechanisms without doing any GC.  This could be explained a lot more clearly in the design.
9. @rptb1 Not only does the design talk about live objects, it not clear how the non-GC does anything about it.
10. @rptb1 m The design doesn't mention that the client code is called on padding objects.  An anti-requirement?
11. @thejayps NI: Comments on the first line of a paragraph hide the paragraph.  Would be better to comment at the end of a paragaph.
12. @UNAA008 NI: The review issue classification is poorly presented for reference during checking.
13. @UNAA008 Nm: There's a relationship between use cases and requirements and none of those are traceable to an origin in the documentation.  @rptb1 There is a link in the references section but it's not very clear or indeed cited from the requirements.
14. @rptb1 NM: We are missing a high-level requirements document.
15. End time 17:15.
16. Logging took 1h.
17. Deferring brainstorm due to pervasive fatigue.
18. How did it happen that I underestimate the checking time?

rptb1:
Executing [proc.review.plan](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#52-review-planning)

1. Start time 15:44.
2. This is a second round of planning because we ran out of time last week. See <https://github.com/Ravenbrook/mps/pull/34#issuecomment-1452233436>.
3. @thejayps requests time to do a "walkthrough" as a pair to discuss what makes this code suitable/unsuitable for Configura's requirements and how to assess its correctness.  We discussed whether that was a part of the review process.  Decided to schedule that separately.
4. Otherwise we think the same proc.review.plan.tactics apply.
5. The checking time was off because of the complexity/subtlety of this change, so estimating by lines of code didn't work too well.  We think we're about half-way through in elapsed time, so another 2-3h should do it.  That's within the bounds of proc.review.plan.time.
6. We'll attack this on Thursday afternoon in the big review time slot.
7. Same roles as before more-or-less.
8. We could consider the walkthrough mentioned above proc.review.plan.homework.
9. @UNAA008 is invited, but optionally. We think his role is mostly fulfilled and he's kinda busy.
10. End time 15:58.  Planning took 14 minutes.

rptb1:
Executing [proc.review.kickoff](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#53-review-kickoff)

1. Start time 13:58
2. @rptb1 focus on correctness, @thejayps focus on consistency, particularly between manual, design, code, test cases.
3. Logging meeting at 15:00.
4. End time 14:08.
5. Kickoff took 10 minutes.

thejayps (commented):
1M 1m 2c 1q:
M: numbered tests eg 104.c and 97.c which were changed in this merge are not documented in the manual or source code index. How does a developer know that they might need to change an arbitrarily named piece of test source when in order to test new functionality? This introduces a risk that tests become out of date and fail to find defects. It also presents a danger in that reviewers can't easily identify whether the appropriate test files have been modified when changes are reviewed.

m: without considering the numbered tests, test coverage for pool_walk is introduced in amcss, amcssth and sncss. Why were these chosen? There may be defects for other pools and configurations that don't test pool_walk, even though the intention for the pool_walk functionality is to make it as far as possible pool-agnostic. If this decision was documented, I haven't yet found the documentation.

comment: although it appears the intention is for mps_pool_walk to replace mps_arena_formatted_objects_walk, which is described as deprecated in the merge, test coverage for the latter is preserved where it exists.

rptb1 (commented):
Executing [proc.review.check](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#54-review-checking)

1. Start time 14:09.
2. End time 14:58
3. 11 minor, 3 comments

rptb1:
Executing [proc.review.log](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#55-review-logging)

1. Start time 15:07.
2. @thejayps 1M, 1m, 2C, 1Q. @rptb1 11m, 4C.
3. 

> How does a developer know that they might need to change an arbitrarily named piece of test source when in order to test new functionality?

M. Similarly with the stress and coverage tests in code, though they are a bit more proximate.

design/object-debug.txt
```diff
@@ -343,11 +343,10 @@ _`.interface.tags.alloc`: Two functions to extend the existing
 ``mps_alloc()`` (request.???.??? proposes to remove the varargs)
 
 ``void (*mps_objects_step_t)(mps_addr_t addr, size_t size, mps_fmt_t format, mps_pool_t pool, void *tag_data, void *p)``
-``void mps_pool_walk(mps_arena_t arena, mps_pool_t pool, mps_objects_step_t step, void *p)``
```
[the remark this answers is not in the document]
rptb1: I think the diffs might be misleading in some way.  Needs investigation.  How could this be deleted before it existed?

rptb1 (commented):
Weird extra review.

code/walk.c
```diff
@@ -381,6 +381,123 @@ void mps_arena_roots_walk(mps_arena_t mps_arena, mps_roots_stepper_t f,
...
+  ScanStateInit(&ss, ts, arena, RankEXACT, trace->white);
+  ss.formatScan = poolWalkScan;
+  ss.areaScan = area_scan;
+  ss.areaScanClosure = closure;
+  ss.fix = walkNoFix;
+
+  RING_FOR(node, &pool->segRing, nextNode) {
+    Bool wasTotal;
+    Seg seg = SegOfPoolRing(node);
+    Bool needSummary = SegRankSet(seg) != RankSetEMPTY;
```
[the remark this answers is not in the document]
rptb1: Fix: Clarified with comments in 07defbe5402fdbf119b70ee83ea420b6b7380131

design/walk.txt
```diff
@@ -0,0 +1,157 @@
...
+_`.intro`: This is the design of the formatted objects walk interface.
+The intended audience is MPS developers.
+
+_`.source`: Based on [GDR_2020-08-30]_.
+
+
+Use cases
+---------
+
+_`.case.reload`: A language runtime that offers hot reloading of code
```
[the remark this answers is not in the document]
rptb1: Fix: Added references and quotes in cd45efb72614a8b42b626d861daa252e29291cda

PR #35, opened 2020-08-31
Remove mps_tramp, deprecated since release 1.111.0

gareth-rees:


UNAA008 (approved):
I walked through this with Gareth on 2020-09-13 and have worked through some of it since.
I have built the MPS from this branch and run the "testall" target of the NT X64 build under NT10 version 1909 build 18363.1016 with success.
These changes are low risk except for the possibility that long standing MPS users may have the inconvenience of removing any lurking calls to the trampoline code. 
I'm approving this pull request with the usual caveat that I am not an MPS or C code expert.

PR #37, opened 2020-09-01
Remove reservoir functions, deprecated since release 1.115.0

gareth-rees:


UNAA008 (approved):
Gareth walked me through this set of changes on 2020-09-13 I've re-examined them since.
I've built X64 and run testall from this branch on an AMD GX-420 CA CPU running NT10 Version 1909 build 18363.1016.
With the usual caveats about my level of expertise I'm approving this PR.

PR #39, opened 2020-09-01
Remove mps_fix, deprecated since release 1.111.0

gareth-rees:


UNAA008 (commented):
I see there is an MPS_FIX macro defined in mps.h, translating to MPS_FIX12 and commented as deprecated.
Should this be removed too?  It does not appear to be documented.

gareth-rees:
I don't really have a strong opinion about `MPS_FIX`. It's just an undocumented alias for `MPS_FIX12`, so it doesn't cost us anything to leave it there (no need for test coverage, or documentation), and it's not a bug magnet like `mps_fix` (because you can call it from inside `MPS_SCAN_BEGIN` without the need for `MPS_FIX_CALL`).

Also, `MPS_FIX` is parallel to `TRACE_FIX` so if we remove the former we ought to rename the latter to `TRACE_FIX12` to maintain the parallelism.

UNAA008 (approved):
Gareth and I worked through this change on 2020-09-13 and subsequently Gareth made the additional change to replace TRACE_FIX in mpm.h during our discussion on 2020-09-27.  The branch tests continue to pass (under W3I6MV) and I'm approving.

PR #40, opened 2020-09-02
Remove mps_alloc_v, deprecated since release 1.112.0.

gareth-rees:
This function was never documented (and never did anything useful), so I haven't put anything in the release notes.


UNAA008 (approved):
Gareth and I discussed this on 2020-09-13.
As this function was never documented and is not apparently useful it seems like a good idea and low risk to remove it.

PR #42, opened 2020-09-27
Remove mps_telemetry_control, deprecated since release 1.111.0

gareth-rees:


UNAA008 (commented):
I think manual/source/topic/security.rst should be modified as part of this change.

gareth-rees:
> I think manual/source/topic/security.rst should be modified as part of this change.

Can you say what changes you would like?

UNAA008:
> 
> 
> > I think manual/source/topic/security.rst should be modified as part of this change.
> 
> Can you say what changes you would like?

Apologies, I was in error.  My case-insensitive brain conflated "mps_telemetry_control" with "MPS_TELEMETRY_CONTROL".  I guess the security issue remains unchanged.

rptb1 (approved):
Looks good to me.

PR #56, opened 2021-01-10
Document the reason for enabling so many warnings.

### LINKED ISSUE #53: Explanation needed for why we turn on so many warnings
The compiler makefile fragments (gc.gmk, ll.gmk, mv.nmk) turn on many warnings along with turning warnings into errors (`-Werror` for GCC and Clang; `/WX` for Microsoft Visual C). There is an important reason why we do this, namely that we recommend [Optimizing for your object format](https://www.ravenbrook.com/project/mps/master/manual/html/guide/build.html#optimizing-for-your-object-format):

> If you are using your own object format, you will also get improved performance by allowing the compiler to do global optimizations between it and the MPS. So if your format implementation is in, say, myformat.c, then you could make a file mymps.c containing:
> ```
> #include "mps.c"
> #include "myformat.c"
> ```
> This will get your format code inlined with the MPS garbage collector.

In order for this to work, the MPS must be able to compile *with whatever compilation and warning options the user has turned on*. This is why we turn on many warning options in our own Makefiles, in order to catch and fix warnings before they are encountered by users.

This is an important requirement and so needs a mention in a design document, and should be cross-referenced from the compiler makefile fragments.

gareth-rees:
Cross-reference the design from the implementation in the compiler makefile fragments.

Fixes #53 (Explanation needed for why we turn on so many warnings)

### DIFF
--- a/code/gc.gmk
+++ b/code/gc.gmk
@@ -12,6 +12,8 @@
 CC = gcc
 CFLAGSDEBUG = -O -g3
 CFLAGSOPT = -O2 -g3
+
+# Warnings that might be enabled by clients <design/config/#.warning.impl>.
 CFLAGSCOMPILER := \
 -Waggregate-return \
 -Wall \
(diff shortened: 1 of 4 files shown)


rptb1 (changes requested):
This change is missing an entry in Document History and is significant enough to want one.  Otherwise excellent.

PR #52, opened 2021-01-10
Ensure that the MPS builds with Clang 10.

gareth-rees:
Clang 10 turns on `-Wimplicit-int-float-conversion`, and on 64-bit platforms this issues a warning that implicit conversions to `double` from `Size` and `unsigned long` "may lose precision".

This commit adds casts to `(double)` for all such conversions. The loss of precision is either impossible in practice (because a double can represent all integers up to 2**53, which is about 9 petabytes, well beyond the addressing capabilities for current CPUs), or else acceptable, because we are accumulating an approximate quantity like "collection work" or "fill size" (that has to cope with loss of precision due in any case), or computing a threshold like the spare commit limit where it is acceptable for it to be somewhat approximate.

Fixes #51 (MPS does not build with clang 10 on Ubuntu 20 x86_64).

Issue #51, opened 2021-01-10
MPS does not build with clang 10 on Ubuntu 20 x86_64


rptb1:
The MPS strict warnings cause build to fail under Ubuntu 20 with clang 10.

Fresh Ubuntu 20 install with just `apt-get install clang`.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
$ clang --version
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /bin
$ arch
x86_64
$ CC=clang ./configure
checking for gcc... clang
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether clang accepts -g... yes
checking for clang option to accept ISO C89... none needed
checking for a BSD-compatible install... /bin/install -c
checking how to run the C preprocessor... clang -E
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking whether __clang__ is declared... yes
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking target platform... Linux x86_64
checking for gnumake... no
checking for gmake... no
checking for make... make
checking sqlite3.h usability... no
checking sqlite3.h presence... no
checking for sqlite3.h... no
configure: creating ./config.status
config.status: creating Makefile
config.status: creating example/scheme/Makefile
CONFIGURE/MAKE IS NOT THE BEST WAY TO BUILD THE MPS
-- see <manual/build.txt> for alternative approaches
$ make
make -C code -f lii6ll.gmk EXTRA_TARGETS="mpseventcnv mpseventpy mpseventtxt"
make[1]: Entering directory '/home/rb/mps/code'
make[2]: Entering directory '/home/rb/mps/code'
make[3]: Entering directory '/home/rb/mps/code'
lii6ll: lii6ll/cool/arena.o
[...]

PR #57, opened 2021-01-11
Macros expressing the intention behind casts to double.

gareth-rees:
As requested in review on #52, this pull request adds macros expressing the intention behind the casts to `(double)` required to avoid warnings from Clang 10 with `-Wimplicit-int-float-conversion` on 64-bit platforms.


rptb1:
The build [fails on LII6LL](https://github.com/Ravenbrook/mps/actions/runs/3934445946/jobs/6729191470#step:4:177) with similar problems to #51 that were fixed in #52, so these macros seem to have regressed that fix.

gareth-rees:
Let's just close this.

rptb1:
Er, I was about to merge this. Is there a reason you don't think it's good?

PR #60, opened 2021-01-23
Prevent races between threads in amcssth.

### LINKED ISSUE #59: Rare failure in amcssth in hot variety on Linux
The failure is very rare on x64-64, less than one in 100 runs. Here are 1000 runs of lii6gc/hot/amcssth, with 5 failures:
```
$ for _ in $(seq 1000); do lii6gc/hot/amcssth > /dev/null; done
amcssth: fmtdy.c:115: dylan_wrapper_check: Assertion `(w[WW] & 3) == 0' failed.
Aborted (core dumped)
amcssth: fmtdy.c:729: dylan_fwd: Assertion `(*(mps_word_t *)old & 3) == 0' failed.
Aborted (core dumped)
amcssth: fmtdy.c:117: dylan_wrapper_check: Assertion `ww[WW] == w[WW]' failed.
Aborted (core dumped)
amcssth: fmtdy.c:117: dylan_wrapper_check: Assertion `ww[WW] == w[WW]' failed.
Aborted (core dumped)
amcssth: fmtdy.c:117: dylan_wrapper_check: Assertion `ww[WW] == w[WW]' failed.
Aborted (core dumped)
```
I've also seen the assertion failure here:
```
amcssth: fmtdy.c:104: dylan_wrapper_check: Assertion `w != NULL' failed.
```
Notes about the envelope of the bug:
* Never occurs in the cool variety (not even on arm64)
* Much rarer on macOS (less than one in a thousand runs)
* Much more common on arm64 (once in every two or three runs).

gareth-rees:
The races arise when multiple threads modify `exactRoots`, or an object pointed to by `exactRoots`, without synchronizing. Prevent races by serializing access to `exactRoots`, and the objects pointed to by `exactRoots`, with a mutex. This doesn't give us the maximum amount of parallelism but it is easy to see that it is correct.

Add mutexes to the test threads interface.

Fixes #59 (Rare failure in amcssth)

### DIFF
--- a/code/amcssth.c
+++ b/code/amcssth.c
@@ -43,6 +43,7 @@ static mps_gen_param_s testChain[genCOUNT] = {
 #define objNULL           ((mps_addr_t)MPS_WORD_CONST(0xDECEA5ED))
 
 
+static testthr_mutex_t exactRootsLock;
 static mps_addr_t exactRoots[exactRootsCOUNT];
 static mps_addr_t ambigRoots[ambigRootsCOUNT];
 
@@ -96,6 +97,7 @@ static void churn(mps_ap_t ap, size_t roots_count)
   ++objs;
   r = (size_t)rnd();
   if (r & 1) {
+    testthr_mutex_lock(&exactRootsLock);
     i = (r >> 1) % exactRootsCOUNT;
     if (exactRoots[i] != objNULL)
       cdie(dylan_check(exactRoots[i]), "dying root check");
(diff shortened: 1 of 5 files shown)


rptb1:
I'll copy some text from our internal discussion here for the record.

The MPS is supposed to allow lockless parallel updates to objects as long as the MPS itself sees a consistent object format for scanning.  And of course we expect the mutator to be able to update objects without locking with the MPS.  The purpose of this test it to exercise these requirements, so https://github.com/Ravenbrook/mps/pull/60/commits/bc492d5b99a7f0d8c2a9302caeec0892660e3db0 is not the right way to solve this problem.

Our debugging session today is more promising! :)

rptb1 (changes requested):
I'll copy some text from our internal discussion here for the record.

The MPS is supposed to allow lockless parallel updates to objects as long as the MPS itself sees a consistent object format for scanning.  And of course we expect the mutator to be able to update objects without locking with the MPS.  The purpose of this test it to exercise these requirements, so https://github.com/Ravenbrook/mps/pull/60/commits/bc492d5b99a7f0d8c2a9302caeec0892660e3db0 is not the right way to solve this problem.

Our debugging session today is more promising, revealing local races between updates to, e.g. the roots array.

rptb1:
Abandoned pending a new proposal to deal with https://github.com/Ravenbrook/mps/issues/59.

Issue #61, opened 2021-02-19
Saving errno and GetLastError in ProtSEHFilter on Windows


fstromback:
In issue #10 @gareth-rees noted that signal handlers might modify errno on Linux systems, which could confuse client code.

The Windows variety, while it does not use signal handlers, uses an exception filter that fills a similar function. Inside that function, MPS might call Windows API functions (e.g. VirtualProtect) which might modify the value in GetLastError, at least on failure (I don't know if the value there is reset on success).

As such, I was thinking if a similar saving/restoring of the value in GetLastError is required on Windows as well, potentially also the value in errno.

I have not yet observed any issues arising from this potential problem, so I do not know if it is an issue (either in theory or in practice). I can try to investigate further if this is something you do not already have a definitive answer to.

gareth-rees:
Hi @fstromback, thanks for raising this issue. Do you know for sure that Windows vectored exception handlers need to save and restore the error values? The documentation for [`AddVectoredExceptionHandler`](https://docs.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-addvectoredexceptionhandler) doesn't mention it (unlike in the case of [`sigaction`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigaction.html)) but that's not dispositive. None of the [vectored exception handling examples](https://docs.microsoft.com/en-us/windows/win32/debug/using-a-vectored-exception-handler) call any Windows API functions.

I'm not in a position to investigate this at the moment, so please go ahead and see if you can figure out if vectored exception handlers need to save & restore error values.

fstromback:
Thank you for your reply, @gareth-rees !

I wrote a small test program to inspect the behaviour on Windows:

```c++
#include <Windows.h>
#include <iostream>

using namespace std;

size_t page_size;
void *memory;

LONG WINAPI handle_exception(struct _EXCEPTION_POINTERS *info) {
    EXCEPTION_RECORD *record = info->ExceptionRecord;

    if (record->ExceptionCode == EXCEPTION_ACCESS_VIOLATION) {
        ULONG_PTR addr = record->ExceptionInformation[1];
        ULONG_PTR mem = (ULONG_PTR)memory;

        if (addr >= mem && addr < mem + page_size) {
            DWORD old;
            VirtualProtect(memory, page_size, PAGE_READWRITE, &old);

            // Call that will fail, and thus modify the value from GetLastError()
[...]

PR #62, opened 2021-03-10
Save and restore GetLastError() on Windows.

fstromback:
This is a patch for the problem outlined in issue #61 - that the value in
GetLastError() is not automatically saved and restored when a vectored exception
handler is called. This is solved the same way as errno was handled on POSIX
systems.

This patch does *not* save and restore errno on Windows, since it seems like the
MPS does not call any functions that modify errno on the critical path on
Windows (generally only functions from POSIX do that).

I have tried to update the documentation similar to what was done for the errno
patch for POSIX, but as I am not entirely familiar with the RST format, I may have
made some mistakes there.

### DIFF
--- a/code/protw3.c
+++ b/code/protw3.c
@@ -44,12 +44,19 @@ LONG WINAPI ProtSEHfilter(LPEXCEPTION_POINTERS info)
   AccessSet mode;
   Addr base, limit;
   LONG action;
+  DWORD lastError;
   MutatorContextStruct context;
 
   er = info->ExceptionRecord;
 
   if(er->ExceptionCode != EXCEPTION_ACCESS_VIOLATION)
     return EXCEPTION_CONTINUE_SEARCH;
+
+  /* This is the first point where we call a Windows API function that
+   * might change the last error. There are also no early returns from
+   * this point onwards.
+   */
+  lastError = GetLastError();
 
   MutatorContextInitFault(&context, info);
 
@@ -96,6 +103,9 @@ LONG WINAPI ProtSEHfilter(LPEXCEPTION_POINTERS info)
     /* <code/vmw3.c#assume.not-last> so it can't be our fault. */
     action = EXCEPTION_CONTINUE_SEARCH;
   }
+
+  /* Restore the last error value before returning. */
+  SetLastError(lastError);
 
   return action;
 }
--- a/design/thread-manager.txt
+++ b/design/thread-manager.txt
@@ -78,6 +78,13 @@ blocked in a system call, and the MPS signal handlers updating
 
 .. _GitHub issue #10: https://github.com/ravenbrook/mps/issues/10
 
+_`.req.thread.lasterror`: It would be nice if on Windows systems the
+MPS does not cause system calls in the mutator to update the value
+returned from ``GetLastError()`` when the exception handler is called
+due to a fault. This may cause the MPS to destroy the previous value
+there. (See `GitHub issue #61`_.)
+
+.. _GitHub issue #61: https://github.com/Ravenbrook/mps/issues/61
 
 Design
 ------
@@ -136,6 +143,12 @@ independent thread, should save and restore its value." All MPS
 signals handlers therefore save and restore ``errno`` using the macros
 ``ERRNO_SAVE`` and ``ERRNO_RESTORE``.
 
+_`.sol.thread.lasterror`: The documentation for ``AddVectoredExceptionHandler``
+does not mention ``GetLastError()`` at all, but testing_ the behaviour
+reveals that any value in ``GetLastError()`` is not preserved. Therefore,
+this value is saved using ``LAST_ERROR_SAVE`` and ``LAST_ERROR_RESTORE``.
+
+.. _testing: https://github.com/Ravenbrook/mps/issues/61
 
 Interface
 ---------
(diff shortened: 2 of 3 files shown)


fstromback:
I should mention: I have tried building and running this on Windows (using the hot variety). I have not run it against the tests in tests/, but it works well when I run it with my language that uses the MPS (so at least it passes the smoke-test).

gareth-rees (changes requested):
Thanks for sending the pull request, which looks good. I've made a couple of comments, but we also need some test coverage. If you look at #32 you'll see that I did this by updating test/function/236.c, which was convenient because this test was already exercising thread suspend and restore on POSIX platforms. In your case you'll want to write a new test case, which will mean figuring out how to use the test system! There are some slender usage notes in test/README, otherwise ask if you need help.

code/misc.h
```diff
@@ -245,6 +245,21 @@ typedef const struct SrcIdStruct {
...
+ * These macros must be used after #include "mpswin.h" so that
+ * GetLastError() and SetLastError() are defined.
+ */
+
+#define LAST_ERROR_SAVE BEGIN DWORD _saved_last_error = GetLastError(); BEGIN
+#define LAST_ERROR_RESTORE END; SetLastError(_saved_last_error); END
+
+#endif /* MPS_OS_W3 */
+
+
```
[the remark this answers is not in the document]
fstromback: Good point. I have fixed this.

code/protw3.c
```diff
@@ -51,51 +51,54 @@ LONG WINAPI ProtSEHfilter(LPEXCEPTION_POINTERS info)
...
-  address = er->ExceptionInformation[1];
-
-  base = (Addr)address;
-  limit = AddrAdd(address, sizeof(Addr));
-
-  if(base < limit) {
-    if(ArenaAccess(base, mode, &context))
-      action = EXCEPTION_CONTINUE_EXECUTION;
-    else
+  /* we don't need to save the last error until here */
```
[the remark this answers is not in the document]
fstromback: Good suggestion, I have adapted this.

fstromback:
I have updated the branch now as suggested above.

I have also added a test case for this (in 237.c). I put that in its own commit now, so it is easy to see that the test fails consistently on Windows first, then when the fix is applied, it passes consistently.

It was a bit tricky to get the right conditions for the issue to be visible. The test is written so that it does its best to set up everything, verify that the right conditions are met and then do the test. This way it can reliably determine that it tests the right thing, and can fail if it for some reasons fails to set up the right conditions (the MPS must decide to raise both the read and write barrier somewhere, so we can trigger a scan in the vectored exception handler).

I have described in quite a bit of detail how it works inside the test file. Hopefully this is understandable.

code/protw3.c
```diff
@@ -16,6 +16,17 @@
...
 
 
+/* Save and restore the value in GetLastError on Windows. See
+ * <design/thread-manager#.sol.thread.lasterror>
+ *
+ * These macros must be used after #include "mpswin.h" so that
+ * GetLastError() and SetLastError() are defined.
+ */
+
+#define LAST_ERROR_SAVE BEGIN DWORD _saved_last_error = GetLastError(); BEGIN
```
NickBarnes, line 26: Why use macros?

design/thread-manager.txt
```diff
@@ -78,6 +78,11 @@ blocked in a system call, and the MPS signal handlers updating
 
 .. _GitHub issue #10: https://github.com/ravenbrook/mps/issues/10
 
+_`.req.thread.lasterror`: It would be nice if on Windows systems the
+MPS does not cause system calls in the mutator to update the value
+returned from ``GetLastError()`` when the exception handler is called
+due to a fault. This may cause the MPS to destroy the previous value
+there.
```
NickBarnes, line 85: link to issue 61.

test/function/237.c
```diff
@@ -0,0 +1,264 @@
...
+ * in the nursery generation and in the older generation, increasing
+ * the likelyhood of us finding a page with the barrier active.
+ *
+ * When we find a page with the read and write barrier active, we
+ * store that pointer and set up the test we want to perform. We want
+ * to set up a situation where we can trigger a barrier hit, and that
+ * the MPS will call a system call that fails in the exception handler
+ * when responding to the hit.
+ *
+ * The tool we use here is a separate thread. We create and register
```
NickBarnes, line 56: This is massively cunning. Applause!

NickBarnes (commented):
Looks OK to me. I'd prefer to have straight calls to GetLastError and SetLastError, rather than macros.

code/protw3.c
fstromback: I used the ERRNO_SAVE and ERRNO_RESTORE as a template here initially. But since it is only used in one location at the moment I agree that the macro-free version is better. I will fix that.

fstromback:
Thank you for your feedback @NickBarnes ! I have made the suggested changes (removed the macros and added the link in the documentation).

fstromback:
Hi again,

I just wanted to check in on the status of this PR. I think I have made all requested changes, but I might have missed something (either in the code or regarding licences etc.). If so, please let me know!

gareth-rees (approved):
Sorry for being so dilatory in getting this merged. I've been putting it off because we no longer have any continuous integration coverage on Windows. (This used to live in a Jenkins instance in the Ravenbrook office, but that was taken offline when the office was closed and no-one has had the spoons to get it going again.) But the continuous integration situation is not going to get better by doing nothing, so I recommend merging as-is (or with the typo fixes below) and worrying about the Windows build at the point when (or if) someone is able to get continuous integration going again.

fstromback:
Thank you for the pointers and the update!

I have fixed the issues you pointed to, I hope they are satisfactory.

gareth-rees:
Pull request merged — thank you for the contribution.

PR #65, opened 2021-05-27
Remove mention of #memorypoolsystem IRC channel from the manual.

gareth-rees:


UNAA008 (approved):
I agree that we should remove the IRC channel reference.  The actual changes to do this look sound to me so I approve.

PR #67, opened 2021-05-27
Update the MPS Manual for compatibility with Sphinx 3.

gareth-rees:


rptb1:
> something's wrong with the layout of the result

This also occurs in master at 7fb1ad681064bc0658aee6b4c7adece6684737c3 with Sphinx 6.1.3 so there's no reason to think it was introduced by this work.

I'm able to remove the layout problem by installing Sphinx 4, so let's specify Sphinx < 5 and make a separate issue for the problem.

rptb1:
Just for the record, I tried `tool/bin/pip install 'Sphinx < 3.6'`.  It wouldn't run on my Ubuntu 22, so I can't easily check whether this branch pedantically meets its stated purpose, but I don't think that matters now.

manual/source/extensions/mps/designs.py
```diff
@@ -34,13 +34,22 @@
 
 '''
 
+# Macros that are improperly named (not all-caps).
```
rptb1, line 37: That's a good point.  These macros were inherited from work by Tony and don't fit MPS conventions.  Logged in #122

rptb1 (approved):
These changes look good.

I did the following steps:
1. Checked I could build the manual locally (results and changes are in the pull request thread #67 ).
2. Read the diffs, focussing on multi-line diffs, but sampling others at random.
3. Sampled pages of the built manual that changed at random and checked them against the diffs.
4. Checked manual/html/design/protocol.html because it could be affected by a mistake in the new doc code.  It wasn't.
5. Read the new paragraphs of manual/html/design/doc.html relating to these changes to make sure they explained the new convention.  They did.

I created #121 and #122 as a result of this review but they are not issues with this work.

rptb1:
Executing [pull request merge procedure](https://github.com/Ravenbrook/mps/blob/99ff0d0fac8b2bd21812526da7a86818b979c265/procedure/pull-request-merge.rst).

Checklist results:
2. Branch [solves related problem](https://github.com/Ravenbrook/mps/pull/67#issuecomment-1386811758).
3. No automated test case is feasible for this.  It would need to try to build the manual with an earlier Sphinx and fail.  Unnecessary.

PR #69, opened 2021-06-03
The MPS no longer supports xci3ll (macOS, i386, Clang/LLVM).

### LINKED ISSUE #68: macOS 10.15 (Catalina) no longer builds or runs i386 code
See the [release notes](https://developer.apple.com/documentation/macos-release-notes/macos-catalina-10_15-release-notes):

> macOS frameworks are now thinned for the x86-64 architecture. Apps that execute i386 code now fail with the EBADARCH error code. The remaining stub frameworks are nonfunctional and exist only for compatibility purposes. (51236070)

Trying to build the xci3ll platform now fails with:
```
Undefined symbols for architecture i386:
  "___error", referenced from:
      _VMInit in mps.a(vmix.o)
      _VMMap in mps.a(vmix.o)
  "___memcpy_chk", referenced from:
      _mps_lib_memcpy in mps.a(mpsliban.o)
      _protBuildReply in mps.a(protxc.o)
  "___memset_chk", referenced from:
      _mps_lib_memset in mps.a(mpsliban.o)
  ...
```
and so on. This means that it is inconvenient for us to test this platform, and so we should declare it unsupported to avoid giving a false impression. The code can be left in place for the moment as it does no harm.

gareth-rees:
As of macOS 10.15 (Catalina), this platform is no longer supported by Apple, so it is inconvenient to develop and test. The code is left in place for developers on older versions of macOS.

Fixes #68 (macOS 10.15 (Catalina) no longer builds or runs i386 code).

### DIFF
--- a/manual/build.txt
+++ b/manual/build.txt
@@ -147,7 +147,6 @@ Platform     OS          Architecture    Compiler      Makefile
 ``w3i3mv``   Windows     IA-32           Microsoft C   ``w3i3mv.nmk``
 ``w3i6mv``   Windows     x86-64          Microsoft C   ``w3i6mv.nmk``
 ``xca6ll``   macOS       ARM64           Clang         ``mps.xcodeproj``
-``xci3ll``   macOS       IA-32           Clang         ``mps.xcodeproj``
 ``xci6ll``   macOS       x86-64          Clang         ``mps.xcodeproj``
 ==========   =========   =============   ============  =================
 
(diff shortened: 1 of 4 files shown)


UNAA008 (approved):
Happy to approve.

rptb1:
This change is incomplete.  It left various files for platform XCI3 in the sources, such as prmcxci3.c and xci3gc.gmk.

Issue #75, opened 2022-01-15
Issue with Apple silicon write-xor-execute memory requirements


tynor:
Hello,

I greatly appreciate the addition of support for macOS/aarch64 in MPS. For me, MPS builds successfully (using the `d894a3f90a` commit), but fails when using the VM arena class because MPS requests `PROT_EXEC` when creating its virtual memory mapping:

```
The MPS detected a problem!
build/src/mps/code/vmix.c:180: MPS ASSERTION FAILED: errno == ENOMEM
See the "Assertions" section in the reference manual:
https://www.ravenbrook.com/project/mps/master/manual/html/topic/error.html#assertions
```

I have verified that removing `PROT_EXEC` from the `mmap(2)` call gets past that particular issue.

One of the features of M1-based Macs is hardware enforcement of only having one of `PROT_WRITE` or `PROT_EXEC`. While Apple [provides APIs](https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon) for managing needing `PROT_EXEC` on pages, my question is whether MPS actually needs to execute code generated on the fly in the first place. I have not looked through much of the code, but I did notice MPS also calls `mprotect(2)` with `PROT_EXEC` but the `AccessSet` values indicate it is only concerned with being able to write.

I'd be very interested to see if there is a workaround I'm missing as well, since I imagine tests are passing on Apple silicon for someone.

Thank you very much.

gareth-rees:
Thanks for reporting the problem. You may be the first to test the arm64 support on Apple Silicon! I have only tested it on Linux.

First, if I understand Apple's documentation correctly, you should be able to work around this by turning off Hardened Runtime for the application, or by setting the "[Allow Unsigned Executable Memory Entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-unsigned-executable-memory)". (Let me know if I've misunderstood this.)

Here are some solution ideas:

1. It would be nice to automatically detect the problem and disable the use of `PROT_EXEC`. (This wouldn't affect backwards-compatibility since no-one can have been using the MPS to manage compiled code objects in Hardened Runtime on Apple Silicon yet, precisely because of this issue.)

[...]

PR #77, opened 2022-01-30
Fix PROT_EXEC restrictions on Apple Silicon

Techcable:
Currently, on Apple Silicon chips, calling `mmap(PROT_WRITE | PROT_EXEC)` will fail due to restrictions on marking memory executable & writable at the same time.
    
See apple developer documentation on this new restriction:
https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon
    
This workaround simply stops requesting PROT_EXEC entirely on Apple Silicon. This is not a complete solution, because it does not allow clients to use seamlessly garbage collection to manage their executable memory.
    
However, it does get basic support working. IT works fine as long as you don't use the MPS to directly allocate executable memory.

See also issue #75 (which is not fully resolved by this)
    
Note that any "proper" solution would still require signficant work to interact with the new Apple APIs for JIT.
This means that it is almost certainly going to require some effort on
the part of the client, and possibly new APIs on the part of the MPS.
    
After this commit, the example scheme interpreter builds and runs correctly on my Apple M1 :)

Note the `./configure` script still doesn't work correctly on Apple Silicon, but I figure that's an issue for another PR 😉

### DIFF
--- a/code/protix.c
+++ b/code/protix.c
@@ -73,16 +73,34 @@ void ProtSet(Addr base, Addr limit, AccessSet mode)
      allowed, possibly that means other operations (read) are also
      allowed.
    */
+
+  /**
+   * NOTE: Apple Silicon doesn't have support for specifying
+   * PROT_EXEC along with writable memory.
+   *
+   * We disable this unconditionally on that platform,
+   * pending a more proper solution.
+   *
+   * See apple developer documentation for more details:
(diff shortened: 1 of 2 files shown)


rptb1:
@Techcable Can you say whether #93 meets your requirements?

thejayps:
JPH + @rptb1: we believe the issue related to this PR is resolved by PR #93. Raise an issue if not the case.

PR #82, opened 2022-12-22
Handle errors from mmap and mprotect due to Apple Hardened Runtime.

### LINKED ISSUE #75: Issue with Apple silicon write-xor-execute memory requirements

gareth-rees:
On Apple Silicon, when [Hardened Runtime](https://developer.apple.com/documentation/security/hardened_runtime) is in force and the application lacks the "[Allow Unsigned Executable Memory Entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-unsigned-executable-memory)", `mmap` and `mprotect` return `EACCES` when an attempt is made to create or modify memory so that it is simultaneously writable and executable.

This commit handles these cases by retrying the operation without the `PROT_EXEC` flag, and setting global variables so that future operations omit the flag.

Work towards #75 (Issue with Apple silicon write-xor-execute memory requirements)

Note that this doesn't completely address #75 because there is no provision for `MAP_JIT`. However, in the absence of a clear customer requirement it's difficult for me to pick a good solution. This commit is a self-contained change that allows developers to work with the MPS on Apple Silicon without having to configure the Allow Unsigned Executable Memory Entitlement.

### DIFF
--- a/code/config.h
+++ b/code/config.h
@@ -714,7 +714,7 @@
  *
  * The MAYBE_HARDENED_RUNTIME macro is true if Apple's "Hardened
  * Runtime" feature may be enabled, and so calls to mmap() and
- * mprotect() with PROT_WRITE | PROT_EXEC will fail with EACCES.
+ * mprotect() with PROT_WRITE | PROT_EXEC may fail with EACCES.
  * See <design/prot#impl.xc.prot.exec> for details.
  */
 #if defined(MPS_OS_XC) && defined(MPS_ARCH_A6)
--- a/code/config.h
+++ b/code/config.h
@@ -710,6 +710,23 @@
 #define WB_DEFER_HIT   1  /* boring scans after barrier hit */
 
 
+/* Apple Hardened Runtime
+ *
+ * .hardened-runtime: On Apple Silicon, applications may be compiled
+ * with Hardened Runtime enabled. These applications have restricted
+ * capabilities: in particular, unless the "Allow Unsigned Executable
+ * Memory Entitlement" is enabled, these applications cannot create
+ * memory that is simultaneously writable and executable. Attempts to
+ * do so using mmap() and mprotect() fail with EACCES.
+ *
+ * See <https://developer.apple.com/documentation/security/hardened_runtime>
+ */
+#if defined(MPS_OS_XC) && defined(MPS_ARCH_A6)
+#define MAYBE_HARDENED_RUNTIME 1
+#else
+#define MAYBE_HARDENED_RUNTIME 0
+#endif
+
 #endif /* config_h */
 
 
--- a/code/config.h
+++ b/code/config.h
@@ -712,14 +712,10 @@
 
 /* Apple Hardened Runtime
  *
- * .hardened-runtime: On Apple Silicon, applications may be compiled
- * with Hardened Runtime enabled. These applications have restricted
- * capabilities: in particular, unless the "Allow Unsigned Executable
- * Memory Entitlement" is enabled, these applications cannot create
- * memory that is simultaneously writable and executable. Attempts to
- * do so using mmap() and mprotect() fail with EACCES.
- *
- * See <https://developer.apple.com/documentation/security/hardened_runtime>
+ * The MAYBE_HARDENED_RUNTIME macro is true if Apple's "Hardened
+ * Runtime" feature may be enabled, and so calls to mmap() and
+ * mprotect() with PROT_WRITE | PROT_EXEC will fail with EACCES.
+ * See <design/prot#impl.xc.prot.exec> for details.
  */
 #if defined(MPS_OS_XC) && defined(MPS_ARCH_A6)
 #define MAYBE_HARDENED_RUNTIME 1
--- a/code/protix.c
+++ b/code/protix.c
@@ -51,7 +51,11 @@
 SRCID(protix, "$Id$");
 
 
-/* Value for memory protection corresponding to AccessSetEMPTY. */
+/* Value for memory protection corresponding to AccessSetEMPTY.
+ * See .convert.access for an explanation of the conversion.
+ * We use a global variable and not a constant so that we can clear
+ * the executable flag from future requests if Apple Hardened Runtime
+ * is detected. See <design/prot#impl.xc.prot.exec> for details. */
 
 static sig_atomic_t prot_all = PROT_READ | PROT_WRITE | PROT_EXEC;
 
@@ -71,7 +75,7 @@ void ProtSet(Addr base, Addr limit, AccessSet mode)
   AVER(AddrOffset(base, limit) <= INT_MAX);     /* should be redundant */
   AVERT(AccessSet, mode);
 
-  /* Convert between MPS AccessSet and UNIX PROT thingies.
+  /* .convert.access: Convert between MPS AccessSet and UNIX PROT thingies.
      In this function, AccessREAD means protect against read accesses
      (disallow them).  PROT_READ means allow read accesses.  Notice that
      this follows a difference in contract as well as style.  AccessREAD
@@ -102,7 +106,10 @@ void ProtSet(Addr base, Addr limit, AccessSet mode)
   if (MAYBE_HARDENED_RUNTIME && result != 0 && errno == EACCES
       && (flags & PROT_WRITE) && (flags & PROT_EXEC))
   {
-    /* See <config.h#hardened-runtime>. */
+    /* Apple Hardened Runtime is enabled, so that we cannot have
+     * memory that is simultaneously writable and executable. Handle
+     * this by dropping the executable part of the request. See
+     * <design/prot#impl.xc.prot.exec> for details. */
     prot_all = PROT_READ | PROT_WRITE;
     result = mprotect((void *)base, (size_t)AddrOffset(base, limit), flags & prot_all);
   }
--- a/code/protix.c
+++ b/code/protix.c
@@ -41,12 +41,20 @@
 
 #include "vm.h"
 
+#include <errno.h>
 #include <limits.h>
+#include <signal.h> /* sig_atomic_t */
 #include <stddef.h>
 #include <sys/mman.h>
 #include <sys/types.h>
 
 SRCID(protix, "$Id$");
+
+
+/* Value for memory protection corresponding to AccessSetEMPTY. */
+
+static sig_atomic_t prot_all = PROT_READ | PROT_WRITE | PROT_EXEC;
+
 
 /* ProtSet -- set protection
  *
@@ -55,7 +63,7 @@ SRCID(protix, "$Id$");
 
 void ProtSet(Addr base, Addr limit, AccessSet mode)
 {
-  int flags;
+  int flags, result;
 
   AVER(sizeof(size_t) == sizeof(Addr));
   AVER(base < limit);
@@ -82,7 +90,7 @@ void ProtSet(Addr base, Addr limit, AccessSet mode)
     flags = PROT_READ | PROT_EXEC;
     break;
   case AccessSetEMPTY:
-    flags = PROT_READ | PROT_WRITE | PROT_EXEC;
+    flags = prot_all;
     break;
   default:
     NOTREACHED;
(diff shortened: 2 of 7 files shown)


code/protix.c
```diff
@@ -41,21 +41,29 @@
...
 #include <limits.h>
+#include <signal.h> /* sig_atomic_t */
 #include <stddef.h>
 #include <sys/mman.h>
 #include <sys/types.h>
 
 SRCID(protix, "$Id$");
 
+
+/* Value for memory protection corresponding to AccessSetEMPTY. */
```
thejayps, line 54: Cross reference to paragraph below re conversion between MPS AccessSet and Unix Prot

code/vmix.c
```diff
@@ -156,11 +157,17 @@ void VMFinish(VM vm)
 }
 
 
+/* Value to use for protection of newly allocated pages. */
```
thejayps, line 160: Cross reference to analagous thing in protix.c and vice versa

code/vmix.c
```diff
@@ -47,6 +47,7 @@
 #include "vm.h"
```
thejayps, line 48: General comment: Some comments from protix.c also apply to corresponding places in this file

thejayps (changes requested):
Joint review with @rptb1, @UNAA008 & @thejayps: 

This fundamentally changes the MPS' behaviour for the user on the Apple platform and this needs to be documented. Code should reference this documentation.

A requirement exists to support dynamically allocated code. If we can't fulfil this requirement on a hardened runtime platform or otherwise we must make sure this is clearly explained

gareth-rees:
> This fundamentally changes the MPS' behaviour for the user on the Apple platform

I think there has been a misunderstanding about the nature of the problem and the purpose of this pull request.

1. The MPS _does not support_ the XCA6LL platform: the latest release is 1.117.0 which does not compile or run on this platform. So there is _no behaviour to change_. This pull request is _one step towards_ implementing support for that platform, so that when you make release 1.118.0 we can announce that it supports XCA6LL.
2. This pull request does not prevent developers on XCA6LL from implementing dynamically allocated code in the MPS. These developers have a couple of other options: they can disable Hardened Runtime or add the Allow Unsigned Executable Memory Entitlement.
3. This pull request enables the MPS to work _in the common case_ where Hardened Runtime is enabled (as it is by default), the Allow Unsigned Executable Memory Entitlement is not configured (as it is not by default) and code does not need to be dynamically allocated (as in many or most programs).

Since this has been misunderstood, I will add some documentation to clarify these points.

rptb1:
> I think there has been a misunderstanding about the nature of the problem and the purpose of this pull request.

I believe we understood everything you say, except the details about Unsigned Executable Memory Entitlement.

The MPS generally supports executable code on the heap.  The majority of MPS deployments rely on it (though not on XCA6LL of course).  If we introduce a platform where that spec is not met, we need to document that we have done that, even if briefly, and ensure that statement is traceable from the code.  I'm sure you're on it.

rptb1:
> The MPS generally supports executable code on the heap.

As far as I can tell (well, as far as I can grep) we don't state this in the manual or design.  This pull request has raised the issue for the first time, because [Apple's Hardened Runtime](https://developer.apple.com/documentation/security/hardened_runtime) is the first platform where it's not trivial.

NickBarnes:
What could support for executable code on the heap look like? Can we sketch that out?
As I understand it, these Apple security changes don’t entirely prevent it, they simply outlaw W&X. So could we provide an API to do W -> X ? What might that look like?

NickBarnes:
E.g. a pool class in which one allocates W & !X, and may force W -> X on an allocated object. MPS has to abandon (whatever that’s called; I’m rusty) any buffer on the same page, and of course the OS prohibits further writes so wouldn’t work for any language runtime or implementation which likes to write to fields in code objects, but presumably this is the sort of thing that language implementors are doing on W^X platforms.
Or am I completely misunderstanding?

rptb1:
> What could support for executable code on the heap look like? Can we sketch that out?

I'm not sure I understand this question.  Configura, for example, allocate objects on the heap in the AMC pool containing code compiled on-the-fly, which they then call.  Any JIT would need it.  (I'm not sure how Apple's system copes with JITs etc.)

rptb1:
See https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit and specifically `MAP_JIT`.  I can't read all of this now, but the requirement for executable memory raises some issues that might affect this code.  In particular, this change uses global static variables to quietly and irreversibly change the nature of allocated memory.  Should the MPS have a proper interface declaring the intention to do JIT-like stuff on the heap?  Can restrictions change at run-time?  Other platforms might start introducing similar restrictions.

thejayps:
I understood NB's question to be assuming the hardened runtime scenario or
similar...

code/protix.c
```diff
@@ -82,15 +90,23 @@ void ProtSet(Addr base, Addr limit, AccessSet mode)
...
     flags = PROT_NONE;
   }
 
   /* .assume.mprotect.base */
-  if(mprotect((void *)base, (size_t)AddrOffset(base, limit), flags) != 0)
+  result = mprotect((void *)base, (size_t)AddrOffset(base, limit), flags);
+  if (MAYBE_HARDENED_RUNTIME && result != 0 && errno == EACCES
+      && (flags & PROT_WRITE) && (flags & PROT_EXEC))
+  {
+    /* See <config.h#hardened-runtime>. */
```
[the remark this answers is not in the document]
gareth-rees: The code base have 43 instances of a cross-reference with no space and no instances with a space, so for consistency I have gone with the majority.

gareth-rees:
I've addressed most of the review comments in the fixup commit. Still thinking about where to put the user documentation.

NickBarnes:
On 4 Jan 2023, at 18:52, Richard Brooksby wrote:

> > What could support for executable code on the heap look like? Can we 
> sketch that out?
>
> I'm not sure I understand this question.  Configura, for example, 
> allocate objects on the heap in the AMC pool containing code compiled 
> on-the-fly, which they then call.  Any JIT would need it.  (I'm not 
> sure how Apple's system copes with JITs etc.)

I mean, in the context of the Apple Hardened Runtime. There are clearly 
several ways to get chosen byte sequences into memory and execute them 
on W^X systems. For example, one could write code to a file (whether on 
a real filesystem or not - see also memfd_create() on Linux) and then 
map it in with MAP_EXEC. I'm a little curious about how the dynamic 
linker works on macOS with the Hardened Runtime, but I imagine it's 
similar to the way it does on Linux these days. Of course some programs 
will use MAP_JIT (for which they will have the necessary "entitlement" - 
I wonder what was wrong with the word "capability").

I approve of the general rule enforced by this aspect of the Hardened 
Runtime: "executable pages shouldn't be writable", also known as "W^X". 
There are ways of enforcing this general policy on loads of different 
OSes including Linuxes, BSDs, and Windows. So what I'm wondering is 
whether the MPS should offer some sort of support for clients who want 
to allocate objects in the arena which contain code which is 
subsequently executed, even on such W^X platforms.

For example, under the Apple Hardened Runtime, can one map anonymous 
pages for writing, write code to them, and then switch them from 
writable to executable before executing the code? (maybe one could even 
do this switch in the signal handler invoked when the 
not-actually-executable code is called, although that could be really 
messy if happening asynchronously with some other thread writing to the 
same page). If that is possible, as I believe it is on some other W^X 
systems, then a future MPS could potentially offer support for systems 
which want to do it (for example, a pool class which supports this 
behaviour somehow). There are a million details. For instance, you'd 
have to invalidate any buffer allocating on the page (segment?) in 
question.

This wouldn't work for language implementations which like to write to 
code objects which are already executable, but those design choices can, 
and probably should, be revisited (IIRC our LLVM backend does this or 
something like it for UNWIND_INFO, but I'm very rusty about all that).
—
Nick B

design/prot.txt
```diff
@@ -141,6 +153,39 @@ _`.impl.w3`: Windows implementation.
 
 _`.impl.xc`: macOS implementation.
 
+_`.impl.xc.prot.exec`: The approach in `.sol.prot.exec`_ of always making
+memory executable causes a difficulty on macOS on Apple Silicon. On
+this platform, programs may enable `Hardened Runtime
+<https://developer.apple.com/documentation/security/hardened_runtime>`_.
+This feature rejects attempts to map or protect memory so that it is
+simultaneously writable and executable. Moreoverm the feature is
```
gareth-rees, line 161: Typo: "Moreoverm"

code/config.h
```diff
@@ -712,14 +712,10 @@
...
- * with Hardened Runtime enabled. These applications have restricted
- * capabilities: in particular, unless the "Allow Unsigned Executable
- * Memory Entitlement" is enabled, these applications cannot create
- * memory that is simultaneously writable and executable. Attempts to
- * do so using mmap() and mprotect() fail with EACCES.
- *
- * See <https://developer.apple.com/documentation/security/hardened_runtime>
+ * The MAYBE_HARDENED_RUNTIME macro is true if Apple's "Hardened
+ * Runtime" feature may be enabled, and so calls to mmap() and
+ * mprotect() with PROT_WRITE | PROT_EXEC will fail with EACCES.
```
gareth-rees, line 717: "may fail" would be better.

design/prot.txt
```diff
@@ -141,6 +153,39 @@ _`.impl.w3`: Windows implementation.
...
+enabled by default in recent versions of Xcode, so that if you install
```
gareth-rees, line 162: A version (e.g. "as of macOS 13 Ventura") would be more future-proof than "recent".

NickBarnes:
W^X was mainly developed to defend against attackers who are able (due to vulnerabilities) to write to some parts of memory which are executable (and which are then executed), but not able to make arbitrary calls to mmap(). Such attackers can't switch W -> X. This is believed, with some reason, to account for quite a lot of past security vulnerabilities.

gareth-rees:
I added a "Limitations" section to the "Platforms" chapter of the Reference Manual, which seems like a plausible place people might look.

thejayps (commented):
2nd review with @rptb1, @UNAA008 & @thejayps:

We think this all looks good but we are currently unable to approve because we can't build on all target platforms. Watch this space.

rptb1:
The branch attached to this pull request doesn't fit with the MPS's durable branch naming scheme of branch/DATE/TOPIC.  This is an opportunity to figure out how renaming the branch affects the infrastructure (GitHub, Travis, etc.).  Please attempt it and say how you did it.

gareth-rees:
> The branch attached to this pull request doesn't fit with the MPS's durable branch naming scheme of branch/DATE/TOPIC

What branch name do you want?

gareth-rees:
GitHub automatically closed the pull request when I renamed the branch. I made pull request #93 to replace this one.

PR #86, opened 2022-12-27
Ensure that the jump buffer is appropriately aligned.

gareth-rees:
The stack capturing mechanism in `STACK_CONTEXT_BEGIN` puts the `StackContextStruct` on the stack, where it will later be scanned using `TraceScanArea`. The `jumpBuffer` must be allocated on the stack at an address with suitable alignment so that `TraceScanArea` will correctly fix any addresses therein. On platform XCA6LL, `jmp_buf` is declared as an array of `int`, which has 4-byte alignment, but we need it to have 8-byte alignment.


thejayps:
@rptb1 : blocked because we are waiting for Apple Silicon CI to be available. This will come from #145.

PR #87, opened 2022-12-29
Report sizes in grains in MMQA test function/150.c.

gareth-rees:
This makes the test portable to XCA6LL, where the page size (and so the default arena grain size) can be 16 kB.


thejayps:
@rptb1 : blocked because waiting for Apple Silicon CI. See #145

Issue #84, opened 2022-12-25
MMQA test cases fail on xca6ll


gareth-rees:
As of pull request #83, the following test cases fail on xca6ll:

- [ ] conerr/60.c
- [ ] conerr/61.c
- [ ] conerr/62.c
- [ ] conerr/63.c
- [ ] conerr/64.c
- [ ] conerr/65.c
- [ ] conerr/66.c
- [ ] conerr/67.c
- [ ] conerr/68.c
- [ ] function/12.c
- [ ] function/44.c
- [ ] function/45.c
- [ ] function/66.c
- [ ] function/105.c
- [ ] function/125.c
- [ ] function/136.c
- [ ] function/150.c

gareth-rees:
Looking at conerr/60.c, the test case is checking that LO asserts on an unaligned exact reference. The assertion is in `loSegFix`, but the test fails because `loSegFix` is not being called. The unaligned exact reference is in the object pointed to by `q`, which is a local variable in `test` and so ought to be on the stack or in a register and so the object ought to be reachable via the thread root.

Disassembly shows that `q` is in register 19:
```
    0x100001c50 <+496>: bl     0x100002190               ; allocone at myfmt.c:100
    0x100001c54 <+500>: mov    x19, x0
```
so it looks as though we are not scanning register 19.

gareth-rees:
Next, looking at function/150.c. This calls `mps_arena_collect` and checks that the "not condemned size" is at most 4096 bytes. But on xca6ll we get
```
!collect_not_condemned=0
```
for the first two collections, but thereafter
```
!collect_not_condemned=16384
```
The test case creates an AWL pool with weak rank, and in the third collection onwards, none of the AWL segments get condemned.

gareth-rees:
Finally, looking at function/136.c. The problem here is similar to function/150.c: the test case assumes that the grain size is 4 kB, and so if it allocates an 8-byte object in an MVFF pool, it cannot allocate a 4 kB object without extending the pool. However, on XCA6LL, the page size can be 16 kB and so the arena grain size gets rounded up to 16 kB and so after allocating an 8-byte object in an MVFF pool, there remains enough space to allocate a 4 kB object without extending the pool.

PR #88, opened 2022-12-29
Deduce arena grain size in MMQA test function/136.c.

gareth-rees:
This makes the test portable to XCA6LL, where the page size (and so the minimum arena grain size) can be 16 kB.

This fixes one of the test failures in #84.


thejayps:
@rptb1 : blocked because waiting for Apple Silicon CI. See https://github.com/Ravenbrook/mps/pull/145

PR #92, opened 2023-01-06
Fix up array overrun warnings on gcc 11.3.0

### LINKED ISSUE #114: MPS does not build on gcc 11.3.0 due to warnings about potential array overruns
[Text copied from https://github.com/Ravenbrook/mps/pull/92#issue-1522942104 . It was out of place in a pull request.]

gcc 11.3.0 is the default on Ubuntu 22.  The MPS fails to build because gcc warns about potential overruns of keyword argument arrays when a DebugVarargs function passes a potentially shorter array to Varargs function, as in https://github.com/Ravenbrook/mps/blob/e78c6e16735d7f16ef86a7f2f8356791a18c8a6e/code/poolams.c#L762-L767

rptb1:
Intended to resolve #114.

Specifies the shortened array in function prototypes to express intentions.

### DIFF
--- a/code/poolams.c
+++ b/code/poolams.c
@@ -747,7 +747,7 @@ static void AMSSegsDestroy(AMS ams)
 
 /* AMSVarargs -- decode obsolete varargs */
 
-static void AMSVarargs(ArgStruct args[MPS_ARGS_MAX], va_list varargs)
+static void AMSVarargs(ArgStruct args[MPS_ARGS_MAX - 1], va_list varargs)
 {
   args[0].key = MPS_KEY_FORMAT;
   args[0].val.format = va_arg(varargs, Format);
@@ -756,6 +756,7 @@ static void AMSVarargs(ArgStruct args[MPS_ARGS_MAX], va_list varargs)
   args[2].key = MPS_KEY_AMS_SUPPORT_AMBIGUOUS;
   args[2].val.b = va_arg(varargs, Bool);
   args[3].key = MPS_KEY_ARGS_END;
+  AVER(MPS_ARGS_MAX - 1 > 3);
   AVERT(ArgList, args);
 }
 
--- a/code/poolmvff.c
+++ b/code/poolmvff.c
@@ -364,7 +364,7 @@ static Res MVFFBufferFill(Addr *baseReturn, Addr *limitReturn,
 
 /* MVFFVarargs -- decode obsolete varargs */
 
-static void MVFFVarargs(ArgStruct args[MPS_ARGS_MAX], va_list varargs)
+static void MVFFVarargs(ArgStruct args[MPS_ARGS_MAX - 1], va_list varargs)
 {
   args[0].key = MPS_KEY_EXTEND_BY;
   args[0].val.size = va_arg(varargs, Size);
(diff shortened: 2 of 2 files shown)


code/poolams.c
```diff
@@ -756,6 +756,7 @@ static void AMSVarargs(ArgStruct args[MPS_ARGS_MAX], va_list varargs)
   args[2].key = MPS_KEY_AMS_SUPPORT_AMBIGUOUS;
   args[2].val.b = va_arg(varargs, Bool);
   args[3].key = MPS_KEY_ARGS_END;
+  AVER(MPS_ARGS_MAX - 1 > 3);
```
gareth-rees, line 759: 1. I think it would make sense to put the assertion first, before dereferencing `args`, otherwise if `args` is too short we get the undefined behaviour before the assertion that is supposed to prevent it.
2. The assertion seems over-strict as the condition we need is `MPS_ARGS_MAX - 1 >= 3`.

code/poolmvff.c
```diff
@@ -379,6 +379,7 @@ static void MVFFVarargs(ArgStruct args[MPS_ARGS_MAX], va_list varargs)
   args[5].key = MPS_KEY_MVFF_FIRST_FIT;
   args[5].val.b = va_arg(varargs, Bool);
   args[6].key = MPS_KEY_ARGS_END;
+  AVER(MPS_ARGS_MAX - 1 > 6);
```
gareth-rees, line 382: Similar points apply here.

code/poolams.c
rptb1: 1. The idea is to put the assertion next to the line that would need to change with it.  The assertion is static at compile-time, so there's no "before" as such.
2. I'm pretty sure filling in args[0] to args[3] uses four slots.  Have I made an obi-wan error?

gareth-rees: I'm not sure what you mean by "The assertion is static at compile-time". If I change `3` to `999` (say) then I get the assertion failure at runtime, not at compile time
```
$ make -f xca6ll.gmk VARIETY=cool amssshe
xca6ll: xca6ll/cool/amssshe
$ xca6ll/cool/amssshe 
xca6ll/cool/amssshe: randomize(): choosing initial state (v3): 1673032430.
poolams.c:759: MPS ASSERTION FAILED: MPS_ARGS_MAX - 1 > 999

Abort trap: 6
```

gareth-rees: I don't think we have static assertions in the MPS, but if we need them we should add them.

rptb1: > I'm not sure what you mean by "The assertion is static at compile-time".

I mean that the outcome is known to the compiler at compile-time.  We have lots of assertions like this.  See, for example, MPMCheck https://github.com/Ravenbrook/mps/blob/5051593b21c80f12d30437ed4eccf197988db98c/code/mpm.c#L37-L98

rptb1: My point is that in the scenario where this assertion would fire, the code is incorrect at compile time, and the compiler "knows" it.  There are no run-time conditions that *won't* fire the assertion. There's a very slim chance that writing to the arg will cause, e.g. a fault before the assertion fires, but I think there's more of a chance of a coding error by separating the two lines.  That's what I was thinking.

We could write things differently: introduce more symbols, remove the numbers, etc. but it's getting a bit elaborate.

thejayps (approved):
I have no further comments to the existing review comments

code/poolams.c
thejayps: On balance I prefer keeping the AVER near to the line that would cause it to change

rptb1:
Executing [pull request merge procedure](https://github.com/Ravenbrook/mps/blob/fe8851fd1c456fb87eedbd289f2a90f60705c303/procedure/pull-request-merge.rst).

Checklist

3. A test case for this would be whether the build is broken, and that is demonstrated by CI records.  No test suite code is feasible.

PR #105, opened 2023-01-11
Fixing markup on reStructuredText documents

rptb1:


rptb1 (commented):
1. m: GitHub does not recognize licence.
2. m: Build instructions have some Sphinx markup (`:ref:` `:term:`) that GitHub can't interpret.
3. m: Supported platforms might be easier to read as a table.

rptb1:
Executing draft express review procedure (to be linked later).
1. I've called in @UNAA008 
2. @UNAA008 and @rptb1 agree the change is low risk.
3. Improvement suggestion: purpose should include getting stuff done
4. We agree none of the issue should prevent exit.
5. One minute brainstorm: Clarify what explicit document history is for.

UNAA008 (approved):
Approving merge as a result of this review https://github.com/Ravenbrook/mps/pull/105#issuecomment-1412637638

rptb1:
Executing [proc.merge.pull-request](https://github.com/Ravenbrook/mps/blob/de31021a5b4ab10dce530e8c75bdcdcda67eecfe/procedure/pull-request-merge.rst)

1. There is not really an automated test case because it would need to view documents via the GitHub web interface and assess how readable they were.  You potentially could check whether *.rst is in a particular format using e.g. `curl` but maybe we can add a rule to #112 to make sure it works anyway.
2. There's a conflict in .travis.yml but it's trivial and I'll deal with it in merge.

Merge completed in 25 mins.

PR #112, opened 2023-01-14
Check the syntax of reStructuredText files during CI

rptb1:
Intended to fix and prevent #113 .

1. Adds a shell script to check the syntax of our reStructuredText files.
2. Adds a GitHub Actions workflow to run that script as part of CI.
3. Fixes up errors found by the above.

Note that in each case where I changed a tag, I grepped the entire tree for references to that tag.

Test case: What we would be testing is whether developers can introduce errors in reStructuredText to master.  That's not practical to automate.  The new script here only checks whether such errors are present *now*.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@
 # Some branches don't need builds.  Add them here to avoid using build
 # resources and unnecessary build messages.  See
 # <https://docs.travis-ci.com/user/conditions-v1>.
-if: NOT branch IN (branch/2023-01-07/pull-request-merge-procedure)
+if: NOT branch IN (branch/2023-01-13/rst-check)
 
 # The main build matrix for POSIX-like systems.
 language: c # see <https://docs.travis-ci.com/user/languages/c/>.
--- a/design/poolamc.txt
+++ b/design/poolamc.txt
@@ -14,14 +14,20 @@ AMC pool class
    single: pool class; AMC design
 
 
-Introduction
-~~~~~~~~~~~~
+Guide Introduction
+~~~~~~~~~~~~~~~~~~
 
-_`.intro`: This document contains a guide (`.guide`_) to the MPS AMC
+.. The intro and readership tags were found to be duplicated by
+   changelist 182116 / commit e9841d23a but not referenced.  But that
+   was just a consequence of two documents being smushed together by
+   RHSK in changelist 168424 / commit b0433b3e9: a guide and a design.
+   It would be good to sort that out.  RB 2023-01-14
+
+_`.guide.intro`: This document contains a guide (`.guide`_) to the MPS AMC
 pool class, followed by the historical initial design
 (`.initial-design`_).
 
-_`.readership`: Any MPS developer.
+_`.guide.readership`: Any MPS developer.
 
 
 Guide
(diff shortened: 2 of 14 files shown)


.travis.yml
```diff
@@ -4,7 +4,7 @@
 # Some branches don't need builds.  Add them here to avoid using build
 # resources and unnecessary build messages.  See
 # <https://docs.travis-ci.com/user/conditions-v1>.
-if: NOT branch IN (branch/2023-01-07/pull-request-merge-procedure)
+if: NOT branch IN (branch/2023-01-13/rst-check)
```
UNAA008, line 8: When this is done, there will be a merge conflict and we need to document how that should be addressed.

design/poolamc.txt
```diff
@@ -14,14 +14,20 @@ AMC pool class
...
-Introduction
-~~~~~~~~~~~~
+Guide Introduction
+~~~~~~~~~~~~~~~~~~
+
+.. The intro and readership tags were found to be duplicated by
+   changelist 182116 / commit e9841d23a but not referenced.  But that
+   was just a consequence of two documents being smushed together by
+   RHSK in changelist 168424 / commit b0433b3e9: a guide and a design.
+   It would be good to sort that out.  RB 2023-01-14
```
UNAA008, line 25: What action is implied by this?  Should an Issue be created?

UNAA008 (changes requested):
I would like to resolve the comments raised before we appprove this.

thejayps (changes requested):
Joint review with @UNAA008.

.travis.yml
rptb1: The pull request merge procedure has provision for this situation.

rptb1:
> This PR is unusual in that it combines a change to the checking machinery with consequential changes to source files. It makes sense to do this in this instance but in general we would require separate Pull Requests for distinct change sets.

You may find that it's not unusual. I've done the same thing every time I introduced checking. Otherwise we break master with the checks.  So I'm not sure I agree.

design/poolamc.txt
rptb1: Created #128 and linked in 7c6b6aad3

rptb1:
> I would like to resolve the comments raised before we appprove this.

Waiting

rptb1:
> I would like to resolve the comments raised before we appprove this.

Please take a look and consider executing https://github.com/Ravenbrook/mps/blob/branch/2023-01-07/pull-request-merge-procedure/procedure/pull-request-merge.rst . Let me know whether what you decide to do.

UNAA008 (approved):
>I would like to resolve the comments raised before we appprove this.

I think this is now ready to merge.

UNAA008:
Executing [proc.merge.pull-request] using
https://github.com/Ravenbrook/mps/blob/40f23128c2bcddd17aadc0c4c3369e6473c754a1/procedure/pull-request-merge.rst

3. There is no automated test case to demonstrate the problem being fixed.  To make one would involve introducing errors purely for demonstration.

It took about an hour get to starting the actual merge procedure, which included time spent on process improvement for the documentation in section 4.  The prerequisite steps were carried out previously. 
Stalled at step 3 of the merge procedure ('git pull --ff-only perforce master') due to authentication difficulties connected with PNJ's idiosyncratic Windows practises.
About 50 minutes was spent investigating those issues and the merge is suspended at this point pending further research.

UNAA008:
Merge procedure recommenced at 16:42.
Merge committed at 17:57 (1h15 after the procedure restarted)
Total time to this point 2h05.
Then an additional 20 minutes working on step 9 and the outcome of the merge.

PR #116, opened 2023-01-14
Enable MMQA tests in CI

### LINKED ISSUE #106: Travis CI does not exercise the MMQA test suite
The problems in issue [MMQA test cases fail on xca6ll](https://github.com/Ravenbrook/mps/issues/84) would likely have been discovered much sooner if the MMQA tests were run by CI.

We should investigate this possibility.

rptb1:
Attempting to resolve #106

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,0 +1,27 @@
+# .github/workflows/c-cpp.yml
+#
+# GitHub CI <https://docs.github.com/en/actions/automating-builds-and-tests/about-continuous-integration>
+#
+# Adapted from <https://github.com/actions/starter-workflows/blob/9f245d9aba830ad16a097a45c78331a05114d815/ci/c-cpp.yml>
+
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
+    branches: [ $default-branch ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: configure
+      run: ./configure
+    - name: make
+      run: make
+    - name: make test
+      run: make test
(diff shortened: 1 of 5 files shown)


rptb1:
This is currently failing to pass test function/136.c https://github.com/Ravenbrook/mps/actions/runs/3939501021/jobs/6739430364#step:6:865 .  Running this locally, I find an assertion failure here https://github.com/Ravenbrook/mps/blob/ed73ba8e96ca534c453f34ea425bc3ff9b103ebd/test/function/136.c#L81-L86

The test reduces the arena commit limit, but seems to assume that it will be able to allocate an 8-byte object in pool2.  This call fails with result code 7, which means it's hitting the commit limit.  The test appears to expect to fail to allocate, but only for the "large objects" that it tries next.  This seems wrong to me, especially if the commit limit is reached exactly by allocating the descriptors for the pool creations.

rptb1:
Possibly related to #88 but trying function/136.c from that branch does not fix the problem here.

rptb1:
> This is currently failing to pass test function/136.c

There seems to be a sensitive boundary condition.  Running
```
rb@plover:~/git/mps/test$ perl test/qa run function/136.c
...
RESULTS
  assert             =  true
  assert_or_abort    =  true
  asserttext         =
    Couldn't allocate first object of size 8 in second pool
  limit_grains       =  39
  seconds            =  0
CONCLUSION
  PASS
```
but running
```
rb@plover:~/git/mps/test$ perl test/qa runset testsets/passing
...
  PASS function/135.c (writeable)
X FAIL function/136.c (writeable)
  PASS function/137.c (writeable)
...
```
with this in log/full.log:
```
RESULTS
  assert             =  true
  assert_or_abort    =  true
  asserttext         =
    Couldn't allocate first object of size 8 in second pool
  limit_grains       =  40
  seconds            =  0
CONCLUSION
  FAIL                  failed on limit_grains: wanted < 40, was 40
```
So it seems that `test/qa run` and `test/qa runset` are not equivalent.

Issue #101, opened 2023-01-10
MPS documentation lacks a review procedure


UNAA008:
The *Memory Pool System pull request merge procedure* Pre-merge checklist mentions code review as an implied requirement for merging but there is no accessible documentation to describe how such reviews should be carried out.

Some of the documentation required will become available when #95 is satisfied but a specific procedure and/or guide is needed to support reviewers properly.

rptb1:
It seems sensible for review logs to appear in GitHub metadata like https://github.com/Ravenbrook/mps/pull/67#pullrequestreview-1253205338

rptb1:
The formal MM group review procedure is at https://info.ravenbrook.com/project/mps/doc/2002-06-18/obsolete-mminfo/mminfo/proc/review/ and to find its adjucts, the kickoff, entry, checking, logging, brainstorm, editing, and exit procedures, see https://info.ravenbrook.com/project/mps/doc/2002-06-18/obsolete-mminfo/mminfo/proc/review?format=raw .

These are internal Ravenbrook documents at the moment. See #95.

PR #117, opened 2023-01-15
Check shell scripts during CI

rptb1:
To help prevent introduction of shell script errors.

Run the [shellcheck](https://www.shellcheck.net/) [linter](https://en.wikipedia.org/wiki/Lint_(software)) on shell scripts in the MPS as part of CI.

Also adds tool/check-shell-scripts for command-line use.

Shell script errors are not known to have caused defects in the MPS, but this was very cheap to implement as a side effect of #112

PR #123, opened 2023-01-19
Adapting Ravenbrook and MM group review procedure to public MPS

rptb1:
Working on #101

### DIFF
--- a/procedure/review.rst
+++ b/procedure/review.rst
@@ -34,14 +34,24 @@ tool.
   - Review is not just (or even mostly) looking at diffs, though
 GitHub encourages this idea.
   - History of MPS review.
-  - Why review is so important to the MPS.]
+  - Why review is so important to the MPS.
+  - Check against book.gilb93.handbook [Gilb_93]_.
+  - Check against kpa.pr, [CMU/SEI-93-TR-025]_, p L3-97
+]
 
 
 A. References
 -------------
 
+.. [CMU/SEU-93-TR-025] "Key Practices of the Capability Maturity
+                       ModelSM, Version 1.1"; Mark C. Paulk,
+                       Charles V. Weber, Suzanne M. Garcia, Mary Beth
+                       Chrissis, Marilyn Bush; Software Engineering
+                       Institute, Carnegie Mellon University; 1993-02;
+                       <https://resources.sei.cmu.edu/asset_files/TechnicalReport/1993_005_001_16214.pdf>.
+
 .. [Gilb_93] "Software Inspection"; Tom Gilb, Dorothy Graham; Addison
-            Wesley; 1993; ISBN 0-201-63181-4; book.gilb93.
+             Wesley; 1993; ISBN 0-201-63181-4; book.gilb93.
 
 
 B. Document History
--- a/procedure/review.rst
+++ b/procedure/review.rst
@@ -1,0 +1,85 @@
+===================================
+Memory Pool System review procedure
+===================================
+
+:tag: proc.review
+:author: Richard Brooksby
+:organization: Ravenbrook Limited
+:date: 2023-01-19
+:confidentiality: public
+:copyright: See `C. Copyright and License`_
+:readership: MPS developers
+:status: draft
+
+
+1. Introduction
+---------------
+
+This document will contain a procedure for reviewing changes to the
+MPS.
+
+This is a placeholder while we arrange to bring in documents from
+outside the MPS Git repository and tree.  For background, see `GitHub
+issue #95 <https://github.com/Ravenbrook/mps/issues/95>`_.
+
+Since this document was created as part of `a project to migrate from
+Perforce to Git (and GitHub)
+<https://github.com/orgs/Ravenbrook/projects/1>`_, this procedure will
+give specifics on conducting review via GitHub.  But the general
+procedure is in no way specific to GitHub, Git, Perforce, or any other
+tool.
+
+
+[Notes for inclusion:
+  - Review is not just (or even mostly) looking at diffs, though
+GitHub encourages this idea.
+  - History of MPS review.
+  - Why review is so important to the MPS.]
+
+
+A. References
+-------------
+
+.. [Gilb_93] "Software Inspection"; Tom Gilb, Dorothy Graham; Addison
+            Wesley; 1993; ISBN 0-201-63181-4; book.gilb93.
+
+
+B. Document History
+-------------------
+
+==========  =====  ==================================================
+2023-01-19  RB_    Created.
+==========  =====  ==================================================
+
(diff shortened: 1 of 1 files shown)


rptb1:
I propose that we do not integrate all the supporting rules, checklists, etc. before reviewing, merging, and adopting this procedure.  I will bring in the most important, based on experience, then let's create an issue for the others and integrate them as needed.  That will get us in to good shape for reviews more quickly, and deal with the most important defect prevention (in the code).

rptb1:
Extra review objectives:

1. Find good ways to use the GitHub review tools.

2. Test whether this is true: https://github.com/Ravenbrook/mps/blob/de6159a75d151f51b8495f6e92d44495224bd49a/procedure/review.rst?plain=1#L425-L429

rptb1:
Notes from test-run of the procedure with @UNAA008 and @thejayps.

1. [proc.review.plan](https://github.com/Ravenbrook/mps/blob/92ab8d819c2c3424963bc1172377afcfef6ea404/procedure/review.rst#52-review-planning) could include a step to ensure coverage of checking roles.  For example, what if nobody understands the language of the thing being checked?  Use this to improve what "Determine" means here https://github.com/Ravenbrook/mps/blob/92ab8d819c2c3424963bc1172377afcfef6ea404/procedure/review.rst?plain=1#L399-L400

2. During review, it's OK to add comments using the diff thing, this will prompt for a new review. Start one. Don't submit review until proc.review.log.  Add your proc.review.check.record to the "Finish your review" comment box.  Don't press "Approve" -- that's the job of the review exit.  Press "Request Changes" if you have found more than zero issues [the author can't do this!].  Otherwise press "Comment".

3. During logging, people can comment anywhere that will appear in "the log" (i.e. the pull request conversation), including starting another review if they really want to.  All issues found in that will be in class.new.  Perhaps suggest that people reply to the issue that made them think of another issue (unless it's somewhere else).

4. proc.review.log.plan says "Slow down if many new issues are being found."  How to tell?

5. The brainstorm was very handy.  Incorporate https://github.com/Ravenbrook/mps/pull/117#issuecomment-1408640480 here.

6. proc.review.pi.record is missing.

7. Where proc.review.pi.action says "improvement suggestion" it's talking about the results of brainstorm.  Focus on that.

8. In proc.review.pi @thejayps wonders in what order he should be doing things and writing things down.  The procedure says "should receive response" which is unclear.  Some advice on when to proc.review.pi.raise (too big, someone should do it, needs management, etc.)

9. A very important principle is that nothing in the output of review is just "lost" or left dangling.  There's always an audit to trail to find out who did what about everything, even if they decided to do nothing.  Why?  Because someone in many years' time might be trying to discover how a defect came to be in the product, whether review failed, who decided what, etc.

rptb1:
>     1. [proc.review.plan](https://github.com/Ravenbrook/mps/blob/92ab8d819c2c3424963bc1172377afcfef6ea404/procedure/review.rst#52-review-planning) could include a step to ensure coverage of checking roles.

Fixed in https://github.com/Ravenbrook/mps/pull/123/commits/29d7190ac1e41e5b0184bb5843c9d59cee3de1b1

>     2. During review, it's OK to add comments using the diff thing,  ...

Fixed in https://github.com/Ravenbrook/mps/pull/123/commits/94fc54a52a95fd965c840f5bd750996100fc657d

>     3. During logging, people can comment anywhere that will appear in "the log" ...

Fixed in https://github.com/Ravenbrook/mps/pull/123/commits/49bb09bffb69b6140e665ff6eb396a4a9f395f1d

>     4. proc.review.log.plan says "Slow down if many new issues are being found."  How to tell?

Fixed in https://github.com/Ravenbrook/mps/pull/123/commits/b2101347e42868e1aedf926ccdc87ec7f136e03f

rptb1:
> 6. proc.review.pi.record is missing.

Fixed in https://github.com/Ravenbrook/mps/pull/123/commits/da35acea891dd41180e717c7f40b865408f30c07

> 7. Where proc.review.pi.action says "improvement suggestion" it's talking about the results of brainstorm.  Focus on that.

Clarified in https://github.com/Ravenbrook/mps/pull/123/commits/da35acea891dd41180e717c7f40b865408f30c07
 
> 8. In proc.review.pi @thejayps wonders in what order he should be doing things and writing things down.  The procedure says "should receive response" which is unclear.  Some advice on when to proc.review.pi.raise (too big, someone should do it, needs management, etc.)

Fixed in https://github.com/Ravenbrook/mps/pull/123/commits/da35acea891dd41180e717c7f40b865408f30c07

rptb1:
The second test run of procedure at https://github.com/Ravenbrook/mps/pull/83#issuecomment-1411256195 went smoothly.  No real problems found in the procedure, but a few good suggestions.

Note for inclusion: Base kickoff, check,log, and brainstorm time seems to be about 2h elapsed for the team we have.  The procedure itself suggests a limit of 3.5h.  Express review is capped at 30m.

rptb1:
Consider renaming "review" to "inspection" for this particular process:
- it's more accurate
- it doesn't co-opt the word "review" which is useful for lots of other things
- it's less likely to confuse people used to other kinds of review
- it makes the point to outsiders that we're doing something more than the usual review
- it might break references from older material a bit, including the old MM stuff.

rptb1:
Some thoughts about how to apply this document to itself, when it comes to it. It's about 3600 lines of text, so that's 6h at the baseline 10 lines/min.  However, it's quite reviewable in sections, especially since most of the procedures a made to be done separately.

Consistency with the introduction and rationale is important.  And the overall structure of the document for its purpose will need attention, possibly both at the start (to see how it appeals to someone new approaching it) and at the end (when everyone's read it all).

We can also switch tactics after one or two sessions.

rptb1:
In https://github.com/Ravenbrook/mps/pull/159#pullrequestreview-1311363100 (I think) @UNAA008 mentioned (but didn't write down) [design/guide.review.txt](https://github.com/Ravenbrook/mps/blob/142999bd7b0d265ea793817d662047953e4e13ab/design/guide.review.txt), [guide.impl.c.format](https://github.com/Ravenbrook/mps/blob/142999bd7b0d265ea793817d662047953e4e13ab/design/guide.impl.c.format.txt), and [guide.impl.c.naming](https://github.com/Ravenbrook/mps/blob/142999bd7b0d265ea793817d662047953e4e13ab/design/guide.impl.c.naming.txt) .  These are Ravenbrook-era review documents that need absorbing/merging into rules and checklists.

thejayps:
Who is responsible for assigning non checking roles (eg scribe, pi) and when does this happen?

Currently it is not happening on some reviews

thejayps:
After looking at "something random" in weekly meeting: https://info.ravenbrook.com/rule/ was found as a source document for review of this

Issue #128, opened 2023-01-25
AMC pool class design document is messy


rptb1:
The AMC pool class design document ([design.mps.poolamc](https://github.com/Ravenbrook/mps/blob/081d79ee2299509615b67cdbe8f0bfdfbd93feab/design/poolamc.txt)) is a messy mixture of two documents, and some of the content is out-of-date.  It needs a major revision.

One example of the problems is https://github.com/Ravenbrook/mps/blob/15b3d00e07baf6292f296dd18d573bed9c359a08/design/poolamc.txt?plain=1#L20-L24

A revision of this document would also be of benefit to our new staff, @thejayps and FOSS contributors.

rptb1:
See also https://github.com/Ravenbrook/mps/pull/159#pullrequestreview-1311360637 item 4 and consider generalising this issue to improve all "old design".  In that review @UNAA008 commented that there would be a reluctance to remove old but potentially useful information, so we might need to think of a way of writing about historical designs.

PR #131, opened 2023-01-27
Eliminate unnecessary write barrier on white segments

### LINKED ISSUE #130: Unnecessary write barrier on white segments
The MPS might be maintaining the write barrier on white segments, and so making unnecessary systems calls after a fix, hurting performance.

See ["Write protection on white segments" (e-mail)](https://info.ravenbrook.com/mail/2016/04/01/16-31-45/0/) from me which includes an experimental patch:

```
==== [//info.ravenbrook.com/project/mps/master/code/poolamc.c#90](http://info.ravenbrook.com/project/mps/master/code/poolamc.c#90) (text+k) ====

1681a1682,1690
> /* We don't need to maintain the remembered set on white segments,
> so get rid of it now to avoid unnecessary write protection costs.
> Don't do this for nailed segments, because they go on the grey
> list and the summary can be used to skip them later. TODO: This
> might not be wise for multiple traces, since another trace might
> be able to eliminate this segment from scanning, and we have not
> yet proved that its contents are dead. */
> if (SegRankSet(seg) != RankSetEMPTY)
> SegSetSummary(seg, RefSetUNIV);
```

In the mail I wrote:

> Seems to be worth about 5% on "xci6ll/hot/gcbench -P 0 -x 1234 amc" on Silverbird, and has a similar impact on "scheme-advanced test-leaf.scm" though it's a bit difficult to measure.

There is a branch and review for this visible in the Perforce branch index at https://info.ravenbrook.com/project/mps/branch/ with a review.  That branch is mirrored in GitHub at https://github.com/Ravenbrook/mps/tree/branch/2016-04-04/unprotect-white

rptb1:
Eliminating the unnecessary write barrier on white segments in order to improve performance.

Closes #130 .


rptb1:
There is an existing review for this branch from @gareth-rees at in ["Re: Unprotecting white set branch is ready for review" (e-mail)](https://info.ravenbrook.com/mail/2018/08/03/10-30-40/0/) and a response is required.

rptb1:
> There is an existing review for this branch from @gareth-rees at in ["Re: Unprotecting white set branch is ready for review" (e-mail)](https://info.ravenbrook.com/mail/2018/08/03/10-30-40/0/) and a response is required.

@gareth-rees wrote:

> 1. The comment says, "We don't need to maintain the remembered set on white segments” but I am not confident that I understand why. I guess the reason must be that white AMC segments never get greyened (and so never need to be considered for scanning), but how do we know that is the case?

AMC is mostly-copying, so any objects not pinned down my ambiguous references are moved to new (grey or black) segments, and the memory occupied by the old copies is recycled.  The old copies are never scanned.

If we implement multiple traces, then copes that appear old for one trace might be scanned by another, and we'll have to measure the tradeoff.

I ought to be able to refer you to design.mps.poolamc but #128 .

rptb1:
> There is an existing review for this branch from @gareth-rees at in ["Re: Unprotecting white set branch is ready for review" (e-mail)](https://info.ravenbrook.com/mail/2018/08/03/10-30-40/0/) and a response is required.

@gareth-rees wrote:

> 2. The comment goes on to say, "Don't do this for nailed segments, because they go on the grey list and the summary can be used to skip them later” which appears to contradict the previous sentence. So white AMC segments do get greyened if they are pinned.

Yes, but by the time the commented code is reached this has already happened, because all ambiguous scanning that might pin a segment happens before any exact scanning.  The comment ought to make this clear.

rptb1:
> There is an existing review for this branch from @gareth-rees at in ["Re: Unprotecting white set branch is ready for review" (e-mail)](https://info.ravenbrook.com/mail/2018/08/03/10-30-40/0/) and a response is required.

@gareth-rees wrote:

> 3. The comment says, “Don't do this for nailed segments” but I don’t see how this is achieved. It looks to me as though we could reach this block of code for a nailed segment if we get an exact fix to an object that was already nailed via a previous ambiguous fix.

The test at https://github.com/Ravenbrook/mps/blob/b763a6a0ce7c938df799b2b112a93bc31b7bffbd/code/poolams.c means that the code is not reached when fixing an ambiguous reference.

However you are correct that an exact fix to a nailed segment will reach this code and cause the remembered set to be forgotten.  That's a mistake and I will correct it.

It may be the case that we perform better by forgetting the remembered set in this case, and we should measure both cases.

rptb1:
> There is an existing review for this branch from @gareth-rees at in ["Re: Unprotecting white set branch is ready for review" (e-mail)](https://info.ravenbrook.com/mail/2018/08/03/10-30-40/0/) and a response is required.

@gareth-rees wrote:

>4. The line `if (SegRankSet(seg) != RankSetEMPTY)` should have a comment mentioning AMCZ (as with other tests of this form).

Totally!  Will fix.

rptb1:
Waiting for performance evidence from @thejayps

PR #134, opened 2023-01-28
Experimental cuckoo hash table of white segments

rptb1:
As https://info.ravenbrook.com/project/mps/branch/ says:

> Experimenting with a cuckoo hash table of white segments. ([job003796](https://www.ravenbrook.com/project/mps/issue/job003796/))


thejayps:
Still needs analysis. Could be a solution to #269

Issue #146, opened 2023-02-09
Provide cmake build system?


waywardmonkeys:
A `cmake` build system could possibly replace the `autoconf`-based build system.

This would be able to generate project files for various IDEs, integrate with something like `vcpkg`, generates `commands.json` to make it easier to drive `clang-tidy` and the static analysis tools, etc.

This could be pretty self-contained and would be fewer files than the `autoconf` system. Perhaps it could even replace the main / primary build system!

rptb1:
Background

The MPS primary build system is `cc`. See ["Compiling for production"](https://github.com/Ravenbrook/mps/blob/40f23128c2bcddd17aadc0c4c3369e6473c754a1/manual/build.txt#compiling-for-production).  :smile: 

The reasons for using autoconf are:

1. make it easy for people who expect `./configure && make` as used in many classic FOSS projects
2. to drive CI and automated tools

The configure script outputs a message saying it's not really the thing to use.  It's just a front end to the real "system" which is in GNU Make.  See ["Building the MPS for development"](https://github.com/Ravenbrook/mps/blob/40f23128c2bcddd17aadc0c4c3369e6473c754a1/manual/build.txt#building-the-mps-for-development).

So we could consider replacing GNU Make with cmake.  Autoconf is a red herring.

rptb1:
> a C++ compiler is required to compile cmake ... more difficult to work with some embedded or experimental targets

This brings up a useful requirement to define.

Whatever "make" we're using needs to be available to developers of the MPS on the "supported" platforms.  GNU Make and NMAKE meet that requirement.

But we can ensure things stay easy for developers porting to platforms that don't have many tools.

From the beginning it's been a design principle that the MPS can be built by a command like:
```
cc -o -D<something>... <source>.c...
```
so that we don't start getting tangly nests of dependencies on tools and the MPS stays easy to port.

That in turn means that people should easily be able to cook up some alternative where they need it.

If we introduce a tool like cmake for convenience, keep that build principle, and give a few instructions, then I think things could work out well.

rptb1:
> a design principle that the MPS can be built by a command like

Defined in detail in [design.mps.config.build.cc](https://github.com/Ravenbrook/mps/blob/40f23128c2bcddd17aadc0c4c3369e6473c754a1/design/config.txt#build-cc)

rptb1:
OK, so kind of interesting, but what do we gain?  It seems like a lot of work to switch.

rptb1:
[...]

PR #148, opened 2023-02-10
Experimenting with CMake configuration

rptb1:
See issue #146

### DIFF
--- a/code/.gitignore
+++ b/code/.gitignore
@@ -61,3 +61,7 @@ tags
 *~
 # GNU make dependencies
 */*/*.d
+# CMake output
+CMakeFiles
+CMakeCache.txt
+cmake_install.cmake
--- a/code/CMakeLists.txt
+++ b/code/CMakeLists.txt
@@ -1,0 +1,127 @@
+# code/CMakeLists.txt -- CMake configuration for the MPS
+#
+# This is just an experimental demonstration configuration to see how
+# we might use CMake to replace GNU Make, NMAKE, and Xcodebuild in the
+# MPS.
+#
+# A complete review of all the GNU Make, NMAKE, etc. files and their
+# usage would be required.
+#
+# Invoke like::
+#
+#   cmake -DCMAKE_BUILD_TYPE=Debug . && make VERBOSE=1 clean all
+#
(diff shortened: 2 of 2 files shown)


rptb1:
[TargetArch.cmake](https://github.com/Ravenbrook/mps/blob/2ab9015214ce2e6815a3fff3e73b79cfe39719e2/code/TargetArch.cmake) resembles [mpstd.h](https://github.com/Ravenbrook/mps/blob/branch/2023-02-10/cmake/code/mpstd.h).  Perhaps we could use the latter to achieve the same result, so that we don't duplicate code.

.gitignore
```diff
@@ -30,3 +30,9 @@ test/test/obj
 *~
 .#.*
 core
+# CMake output
+CMakeFiles
+CMakeCache.txt
+cmake_install.cmake
```
waywardmonkeys, line 36: `CMakeFiles`, `CMakeCache.txt` and `cmake_install.cmake` and friends shouldn't be present here since they should only be in build directories. I tend to always build in a separate build directory and it doesn't even have to be under the root mps directory.

CMakeLists.txt
```diff
@@ -0,0 +1,427 @@
...
+target_compile_options(mps PRIVATE ${CFLAGSCOMPILERSTRICT})
+
+add_library(mpsplan STATIC ${PLINTH})
+target_compile_options(mpsplan PRIVATE ${CFLAGSCOMPILERSTRICT})
+
+add_executable(gcbench code/gcbench.c ${FMTDYTST} ${TESTLIB} ${TESTTHR})
+target_compile_options(gcbench PRIVATE ${CFLAGSCOMPILERSTRICT})
+# FIXME: Should be C89 but CMake doesn't know it.
+# FIXME: How to set this for all targets not just gcbench?
+# target_compile_features(gcbench PRIVATE c_std_90)
```
waywardmonkeys, line 408: Towards the top of the file, you can do:

```
SET(CMAKE_C_STANDARD 90)
```

That will set it for all subsequent targets.

mgood7123:
OS and architecture specifics should be detectable at the compiler level, combined with try_compile to test for specifics while configuring cmake

Issue #125, opened 2023-01-19
Version and release procedures use Perforce


rptb1:
The [version create](https://github.com/Ravenbrook/mps/blob/081d79ee2299509615b67cdbe8f0bfdfbd93feab/procedure/version-create.rst), and [release build](https://github.com/Ravenbrook/mps/blob/081d79ee2299509615b67cdbe8f0bfdfbd93feab/procedure/release-build.rst) procedures are specific to Ravenbrook's Perforce.

Ravenbrook needs them updated in order to [migrate to Git and GitHub](https://github.com/orgs/Ravenbrook/projects/1) (see #98 ).

They need to be useful to other FOSS developers of the MPS.

rptb1:
We should also examine the Python programs related to branch and release procedure:
- https://github.com/Ravenbrook/mps/blob/0d071c0af272a5c323cfd2277b2d6518f4e1f1b2/tool/branch
- https://github.com/Ravenbrook/mps/blob/0d071c0af272a5c323cfd2277b2d6518f4e1f1b2/tool/release

Unfortunately, it may be best to delete these if they aren't relevant to Git.

rptb1:
The new procedure must ensure the relevant version of the manual is built, published, and linked.  See #141 .

UNAA008:
The version-create procedure currently assumes that there is an up to date plan or list of requirements to be satisfied by the version branch and that there is a single thread of numbered versions.
There's no mention of other forms of versioning that may happen with a more distributed set of interested parties developing with he MPS as a component.  Issue #126 may be relevant to this, depending on the intended life cycle.
The procedure requires an update to the master version of code/version.c, which would be awkward if there are parallel version branches active.
Arguably we should solve the version threading problem when it actually arises, but we should at least note it.
The planning problem is a separate Issue to be raised somewhere.

UNAA008:
The table of versions is updated by the version-create procedure.
This is currently part of the MPS project tree at Ravenbrook and not curently updateable from outside.

rptb1:
> The version-create procedure currently assumes that there is an up to date plan or list of requirements

We could bring https://www.ravenbrook.com/project/mps/plan/ in to the repo.  See also #95.

rptb1:
Updating the release procedure must also update manual/build.txt and possibly readme.txt.  See https://github.com/Ravenbrook/mps/blob/40f23128c2bcddd17aadc0c4c3369e6473c754a1/manual/build.txt?plain=1#L17-L20

[...]

Issue #95, opened 2023-01-06
Repo needs copies of MPS project and Ravenbrook documents


rptb1:
The MPS is migrating from being "mastered" inside Ravenbrook to living in git (and on GitHub).  Ravenbrook's infrastructure is based on Perforce SCM, where the [current](https://github.com/Ravenbrook/mps/commit/e78c6e16735d7f16ef86a7f2f8356791a18c8a6e) git repository is just a mirror of a "codeline" and its "branches" that form only part of a wider "project" in a company-wide "repository".  As a consequence, the stuff in the git repo refers to important documents and resources that aren't part of it, and sometimes not visible.

(Quite a lot of the MPS "project" part of the Ravenbrook repository is visible at https://www.ravenbrook.com/project/mps/ )

But the style of git (and GitHub) use tends to keep copies and versions of these resources in the git repo alongside the code, and this makes sense especially now the project is FOSS under the BSD licence.

So, we should branch various things from elsewhere in the Ravenbrook repository in to the MPS git repo.  This issue can help to track and discuss how to do it.

A really good example is the code review rules.  Currently, the contributing guidelines https://github.com/Ravenbrook/mps/blob/e78c6e16735d7f16ef86a7f2f8356791a18c8a6e/contributing.rst#review say:
>  There are a fairly large number of rules, both low- and high-level that your code must follow in order to be accepted. These rules are the result of continuous process improvement to prevent defects. Unfortunately, we do not have many of them published at present. We apologise if you find it frustrating that we do not accept your changes as they are.

Our rules are a valuable result of nearly 30 years of high-reliability development of the MPS, and we not only want to share them, but encourage collaboration, including the ability of other people to review and apply them so that the MPS continues to deploy with its extremely low defect rate.

I encourage other team members, past and present, to add more items to this issue.

rptb1:
A technical note about how to bring new content into the repo.  We should not simply copy files.  Those files have history, and that history is valuable data when trying to figure out, for example, what rules were applied when.  Or simply who said what.  Fortunately, there is a way.  We can ask git to merge their histories.

The MPS git repo is exported from Perforce using Perfoce Git Fusion.  Unfortunately, we can't adjust Git Fusion's view to include the history of new files, because that would cause Git Fusion to essentially "rebase" all of MPS history.  That's not really a flaw in Git Fusion, since it was never really designed to maintain a permanent two-way connection between Perforce and git.

[...]

PR #171, opened 2023-02-22
Migrating and updating the project plan from Perforce

rptb1:
Working on #95 .

Solving https://github.com/Ravenbrook/mps/issues/125#issuecomment-1420903293 .


rptb1:
Version 1.117 plan wasn't recorded.  It would not be correct to retrospectively "plan" it, unless we can uncover a plan that was made at the time.  @gareth-rees might have some record elsewhere.

PR #175, opened 2023-02-24
Initial WebAssembly support.

waywardmonkeys:
Fixes #168.

Issue #168, opened 2023-02-22
WebAssembly support


waywardmonkeys:
This is a bug to track the addition of WebAssembly support.

waywardmonkeys:
There's at least 3 potential targets here, I think that MPS should support 2 in the end:

* emscripten
* wasi

In both cases, the CPU is 'WebAssembly' and the compiler is most likely to be clang (llvm).

What should the platform name be for each?

Right now, I have:

```
+#elif defined(__EMSCRIPTEN__)
+#if defined(CONFIG_PF_STRING) && ! defined(CONFIG_PF_EMI3LL)
+#error "specified CONFIG_PF_... inconsistent with detected emi3ll"
+#endif
+#define MPS_PF_EMI3LL
+#define MPS_PF_STRING   "emi3ll"
+#define MPS_OS_EM
+#define MPS_ARCH_I3
+#define MPS_BUILD_LL
+#define MPS_T_WORD      unsigned long
+#define MPS_T_ULONGEST  unsigned long
+#define MPS_WORD_WIDTH  32
+#define MPS_WORD_SHIFT  5
+#define MPS_PF_ALIGN    4
```

(similar to what I had before)

Clearly the `i3` part is wrong ...

Tell me what you want here and I'll carry on with my changes (which so far work at least minimally to run the scheme example).

rptb1:
The architecture code selects architecture-specific code for things like stack scanning and the mutator context.  See for example [prmclii6.c](https://github.com/Ravenbrook/mps/blob/e179634de028b2db090ee6a2358391155102b74f/code/prmclii6.c) . So if there is likely to be more than one similar "WebAssembly" architecture (e.g. a 64-bit variant), then pick a letter for the family and '3' for the 32-bit variant.  Sadly 'W' and 'A' are already taken!  J for Javascript?

You'll need an OS code to select modules with memory mapping, locking, threading primitives.  Again, are they all the same or more than one?  What is the run-time environment called?

Please let me know if anything in [Porting the MPS](https://memory-pool-system.readthedocs.io/en/latest/topic/porting.html) is unclear.

rptb1:
[mps.c](https://github.com/Ravenbrook/mps/blob/e179634de028b2db090ee6a2358391155102b74f/code/mps.c) spells out how source code is selected based on platform code.  You'll be adding sets to that.

rptb1:
[...]

PR #181, opened 2023-02-27
ANSI platform builds shouldn't link pthreads.

waywardmonkeys:
Since this isn't using threads or POSIX locks, the `-lpthreads` should not be needed and having it only makes it easier to accidentally add some dependency on it.

[Edit by @rptb1 : @waywardmonkeys requires this change to build on WebAssembly.  See #168 and #175 .]

### DIFF
--- a/code/anangc.gmk
+++ b/code/anangc.gmk
@@ -16,7 +16,7 @@ MPMPF = \
     than.c \
     vman.c
 
-LIBS = -lm -lpthread
+LIBS = -lm
 
 include gc.gmk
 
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
It looks like the ANSI target is only tested on Travis CI (without digging deeply) and then we run into the same issue as the WebAssembly port where we're building threaded targets (some tests) in a single-threaded build.

thejayps:
JH+RB: essential as this is likely slowing down developers. Awaiting proc.review.express soon

rptb1:
I'll clarify what this is all about with @waywardmonkeys , who's the expert on the WebAssembly platform family.

Issue #165, opened 2023-02-22
memorymanagement.org is build on Ravenbrook servers


rptb1:
As part of migrating the MPS to Git (currently on GitHub) we have moved publication of the MPS manual from Ravenbrook's servers to Read the Docs (see #141 ).

However, a second site, the [Memory Management Reference](https://www.memorymanagement.org/), is also built from the MPS manual sources, and that is still both hosted on Ravenbrook's servers and built by Ravenbrook's automation.

In this case, the domain name memorymangement.org is important and need to be retained.

rptb1:
Read the Docs can publish with a [custom domain](https://docs.readthedocs.io/en/stable/custom-domains.html#custom-domains).

rptb1:
A few notes about how things work.

1. There is a separate [Memory Management Reference project](https://info.ravenbrook.com/project/mmref/) in Ravenbrook's Perforce repository, with its own versions etc.
2. Over time the [master codeline of this project](https://info.ravenbrook.com/project/mmref/master/) evolved into a stub that extracts the Memory Management Reference web site from the MPS manual.
3. This works by first building the MPS manual with the environment variable MMREF set, then processing it with [make-mmref.py](https://info.ravenbrook.com/project/mmref/master/make-mmref.py)
4. In the www LXD container, the [www robot Perforce client](https://info.ravenbrook.com/infosys/cgi/perfbrowse.cgi?%40client+www-www-home) maps the master codeline of the MPS and MMRef projects together to build https://www.memorymanagement.org/ .
5. Like many things, this setup has grown a bit complicated and fragile over time.

rptb1:
Read the Docs [supports custom domains](https://docs.readthedocs.io/en/stable/custom-domains.html#custom-domains) so we could build and host the Memory Management Reference with them.

PR #187, opened 2023-03-03
Build the Memory Management Reference from the MPS tree

rptb1:
Changes to the MPS manual could break the [Memory Management Reference](https://www.memorymanagement.org/).  Incorporating the build will allow us to check it, both manually and as part of CI.

See https://github.com/Ravenbrook/mps/pull/166#issuecomment-1447343865 .

See also #165 though this branch is not intended to resolve that issue, only mitigate risks.

### DIFF
--- a/manual/.gitignore
+++ b/manual/.gitignore
@@ -2,6 +2,7 @@ doctrees
 converted
 epub
 html
+mmref.in
 mmref
 source/design/*.rst
 source/design/*.svg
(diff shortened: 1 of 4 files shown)


rptb1:
Here is a partial build of the Memory Management Reference using Read the Docs from this branch https://mmref.readthedocs.io/en/branch-2023-03-03-make-mmref/

This is achieved by setting the `MMREF` environment variable to 1 in the [Read the Docs configuration](https://readthedocs.org/dashboard/mmref/environmentvariables/).

It's only a partial build because we post-process the Sphinx output to produce https://www.memorymanagement.org/ and Read the Docs is not running that script.  It probably could.

rptb1:
Submitted https://github.com/readthedocs/readthedocs.org/issues/10103

PR #188, opened 2023-03-04
Unix: Use sysconf(_SC_PAGESIZE), not getpagesize()

waywardmonkeys:
The `getpagesize()` is a legacy function and is not available in the default configurations on some platforms.

The documentation for `getpagesize()` recommends using the POSIX `sysconf(_SC_PAGESIZE)` instead.

Additionally, `sysconf(_SC_PAGESIZE)` returns a `long` rather than an `int`, so the code is updated to handle that as well.

Sources:

* https://man7.org/linux/man-pages/man2/getpagesize.2.html
* https://pubs.opengroup.org/onlinepubs/7908799/xsh/getpagesize.html

### DIFF
--- a/code/vmix.c
+++ b/code/vmix.c
@@ -50,7 +50,7 @@
 #include <signal.h> /* sig_atomic_t */
 #include <sys/mman.h> /* see .feature.li in config.h */
 #include <sys/types.h> /* mmap, munmap */
-#include <unistd.h> /* getpagesize */
+#include <unistd.h> /* sysconf, _SC_PAGESIZE */
 
 SRCID(vmix, "$Id$");
 
@@ -59,10 +59,10 @@ SRCID(vmix, "$Id$");
 
 Size PageSize(void)
 {
-  int pageSize;
+  long pageSize;
 
   /* Find out the operating system page size */
-  pageSize = getpagesize();
+  pageSize = sysconf(_SC_PAGESIZE);
 
   /* Check the page size will fit in a Size. */
   AVER((unsigned long)pageSize <= (unsigned long)(Size)-1);
--- a/code/vmix.c
+++ b/code/vmix.c
@@ -61,7 +61,8 @@ Size PageSize(void)
 {
   long pageSize;
 
-  /* Find out the operating system page size */
+  /* Find out the operating system page size
+     (see design.mps.vm.impl.ix.page.size) */
   pageSize = sysconf(_SC_PAGESIZE);
 
   /* Check the page size will fit in a Size. */
--- a/design/vm.txt
+++ b/design/vm.txt
@@ -269,7 +269,7 @@ Unix implementation
 
 _`.impl.ix`: In ``vmix.c``.
 
-_`.impl.ix.page.size`: The page size is given by ``getpagesize()``.
+_`.impl.ix.page.size`: The page size is given by ``sysconf(_SC_PAGESIZE)``.
 
 _`.impl.ix.param`: Decodes no keyword arguments.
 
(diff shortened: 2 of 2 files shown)


rptb1:
> The `getpagesize()` is a legacy function and is not available in the default configurations on some platforms.

Have you seen it missing on a platform yet?  Which ones?

waywardmonkeys:
> Have you seen it missing on a platform yet? Which ones?

On anything using musl libc, like Emscripten or Alpine, it is protected by:

```
#if defined(_GNU_SOURCE) || defined(_BSD_SOURCE)
```

On GNU libc:

```
# if defined __USE_MISC || !defined __USE_XOPEN2K
```

In the macOS 12.1 SDK:

```
/* Removed in Issue 6 */
#if !defined(_POSIX_C_SOURCE) || _POSIX_C_SOURCE < 200112L
```

So it is possible to have it not be defined, with some effort on a variety of platforms.

We currently need `_GNU_SOURCE` enabled on Linux, for `MAP_ANON`, so this isn't yet a critical issue.

thejayps:
JH discussing with RB: this is optional as it is a worthwhile improvement but isn't affect customers in the near future

rptb1 (approved):
Executing [proc.express.check](https://github.com/Ravenbrook/mps/blob/586336a7021c2276a3f798ad10e6e4ab5f438226/procedure/review.rst#12-express-review)

1. Start time 14:30.
2. Reading through [design.mps.vm](https://github.com/Ravenbrook/mps/blob/2191f3c75df254ca272a0ce9e417983d31424a95/design/vm.txt)
3. m. [design.mps.vm.req.reserve.exclusive](https://github.com/Ravenbrook/mps/blob/2191f3c75df254ca272a0ce9e417983d31424a95/design/vm.txt?plain=1#L46-L49) says something unclear and probably wrong about OS support for reserving address space.
4. m. [design.mps.vm.req.reserve.size](https://github.com/Ravenbrook/mps/blob/2191f3c75df254ca272a0ce9e417983d31424a95/design/vm.txt?plain=1#LL57-L58C46) is not clear about *why* this is necessary for zones to work.
5. m. (On the same theme.) [design.mps.vm.req](https://github.com/Ravenbrook/mps/blob/2191f3c75df254ca272a0ce9e417983d31424a95/design/vm.txt?plain=1#L33-L122) was written long ago and does not connect to subsequent designs very well.  Add cross-references.
6. Checking finished 14:44.
7. Checking time 14 mins.
8. Found 3 minor defects.

UNAA008 (commented):
C In https://github.com/Ravenbrook/mps/blob/2191f3c75df254ca272a0ce9e417983d31424a95/code/vmix.c#L69, why does the check cast to unsigned long?
m Copyright date should be updated.

rptb1:
Executing [proc.merge.pull-request](https://github.com/Ravenbrook/mps/blob/3104ab67ae1334f5996a465d881bd713a4427922/procedure/pull-request-merge.rst)

1. Start time 22:14.
2. There is no automated test case.  How could we automatically spot dependencies on deprecated system calls?  Good question.
3. Merge took 5 mins.

rptb1:
> 2\. There is no automated test case.  How could we automatically spot dependencies on deprecated system calls?  Good question.

https://stackoverflow.com/questions/1176121/whats-the-best-way-to-check-if-c-code-has-deprecated-posix-calls says:

> [Lint](http://www.gimpel.com/Discussion.cfm#2) can find deprecated things (-deprecated flag).
> 
> [Coverity](http://www.coverity.com/) finds this kind of issue routinely with c and c++ codebases.

waywardmonkeys:
That would require that the libc headers actually mark it as deprecated ... which is often not the case!

Maybe it is worth proposing this to some libc implementation, but that's a rabbit hole that I'd rather avoid. :)

Issue #190, opened 2023-03-05
MPS needs a Git-specific procedures section


rptb1:
The MPS is migrating from Perforce to git and using GitHub. We will need documentation of Git-specific procedures.  (As opposed to GitHub-specific, see #94 ).

For example, to implement patching of versions branches with fixes applied to master, or vice-versa, we can use the procedure described in keybase://chat/ravenbrook#mps/3065 :

> You can cherry-pick merges, which is one way of applying the same branch to e.g. master and version/1.117.  Let's say we want to apply branch/2023-01-13/rst-check , which has been merged into master, to version/1.117.  You can find the merge with `git log --oneline --first-parent -G branch/2023-01-13/rst-check master` then apply that same merge patch to the current branch with `git cherry-pick -x -m 1 3c62e29ad`.  The `-m 1` extracts a patch against the first parent (previous commit to master) and the `-x` ensures the origin of the cherry-pick is recorded in the commit.
>
> The search works because step 4 of https://github.com/Ravenbrook/mps/blob/0a27a849155fa4396d39ed5204f2d62f0b03ccd1/procedure/pull-request-merge.rst#5-merging-procedure mandates putting the full name of the branch in the commit comment.
>
> It's still not as good as `p4 integrate` but it allows us to implement "Fixing defects" from PQTCM.  See https://www.ravenbrook.com/doc/1999/05/20/pqtcm/ figure 7.

Issue #124, opened 2023-01-19
Version 1.117 does not build from GitHub using CI


rptb1:
The version branch version/1.117 does not build cleanly in the new Git regime.

This is important because we want the capability to create new releases.

And currently, we can't make a new release of MPS 1.117.x from Git and GitHub, so we should sort that out as part of the migration project.

This issue includes #102 .

rptb1:
An alternative is to mothball 1.117 and make 1.118 but I think it would be risky to make a new version of the MPS *during* our migration to Git, since that's introducing a bunch of changes to everything.

rptb1:
The MPS follows the broad procedure in "[Product Quality through Change Management](https://www.ravenbrook.com/doc/1999/05/20/pqtcm/)" but these aren't always easy to implement using Git.  In particular, with Perforce it's very easy to "integrate" (generalized merge) a change branch into two independent branches (e.g. master and version/1.117), but Git doesn't really support that well.

We have to develop Git practices that will allow to achieve the same results.

rptb1:
> into two semi-independent branches (e.g. master and version/1.117),

See https://github.com/Ravenbrook/mps/pull/97#issuecomment-1397086421

rptb1:
> We have to develop Git practices that will allow to achieve the same results (or revise our goals!)

See the procedure in https://github.com/Ravenbrook/mps/issues/190#issue-1610125376 and indeed the whole of #190 .

PR #192, opened 2023-03-08
Fixing CI builds of version 1.117

rptb1:
Working on #124 .

Work plan:
1. Cherry pick CI changes from master using procedure described here https://github.com/Ravenbrook/mps/issues/190#issue-1610125376
2. Fix up build issues, as far as possible by cherry picking from master.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,113 @@
+# .github/workflows/build-and-test.yml -- GitHub CI build and test configuration for the MPS
+#
+# Copyright (c) 2019-2022 `GitHub contributors`_ (MIT License).
+# Copyright (c) 2023 Ravenbrook Limited. See end of file for license.
+#
+# See design.mps.test.ci.
+#
+# TODO: Exclude certain branches.
+#
+# TODO: Regular builds of version branches.  See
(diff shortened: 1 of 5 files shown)


rptb1:
Notes on tactics.

I searched for relevant commits to master using commands like `git log --oneline --first-parent -E --grep '^Merg.*ci' master` then attempted to cherry pick them using commands like `git cherry-pick -x -m 1 1b5ab8313`.  This worked when the same change could be applied.

However, this isn't always the case.  For example, branch/2021-01-10/implicit-conversion can't be applied as-is to version 1.115 because it depends on an earlier branch branch/2016-03-04/spare-fraction, but that branch modifies the behaviour of the MPS and should not be merged into version 1.115.  So some thought will be required to make an equivalent change.

```
rb@plover:~/git/mps$ git log --oneline --first-parent -E -G 'ArenaSpareCommitLimit' master
8797baba7 Merge pull request #52 from Ravenbrook/branch/2021-01-10/implicit-conversion
71da28232 Merge branch/2016-03-04/spare-fraction into the master sources.
0beab5fe3 merging from MMdrj_devel_arena_hysteresis, change.epcore.brisling.160132
```

rptb1:
Manual is up (sans designs) at https://memory-pool-system--192.org.readthedocs.build/en/192/

At this point the macOS build is breaking because Xcode no longer supports i386 (since version 10 I think) and I have limited access to macOS.

PR #198, opened 2023-03-13
Documentation of importance

### LINKED ISSUE #142: MPS priority conventions aren't public
Throughout the MPS documentation, issue database, etc. we use conventions for expression priorities, based on a four-level keyword system: critical, essential, optional, nice.  This corresponds to the [variant MoSCoW method](https://en.wikipedia.org/wiki/MoSCoW_method#Variants) and [RFC 2199](https://www.ietf.org/rfc/inline-errata/rfc2119.html).  We also distinguish "importance" from "urgency".

With #98 and [GitHub migration](https://github.com/orgs/Ravenbrook/projects/1) these keywords are creeping into other places, like [GitHub labels](https://github.com/Ravenbrook/mps/issues/labels).

We need some clear public definition and guidelines so that people:

1. can interpret what documents, issues, etc. mean
2. contribute and communicate priority

See also https://github.com/Ravenbrook/mps/issues/98#issuecomment-1404677166 about migrating the Perforce "jobspec", which relies on this.

rptb1:
Introduces [documentation of importance](https://github.com/Ravenbrook/mps/blob/branch/2023-03-13/importance/procedure/importance.rst).

Fixes #142 .

### DIFF
--- a/procedure/importance.rst
+++ b/procedure/importance.rst
@@ -1,0 +1,190 @@
+=============================
+Importance in the MPS Project
+=============================
+
+:tag: importance
+:author: Richard Brooksby
+:organization: Ravenbrook Limited
+:date: 2023-03-13
+:confidentiality: public
+:copyright: See `C. Copyright and License`_
+:readership: developers, managers
+
+1. Introduction
+===============
+
+This document defines *importance keywords* as used in the Memory Pool
+System project.
+
+
+2. Overview
+============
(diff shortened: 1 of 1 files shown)


rptb1:
When merging this pull request, we should also update the [GitHub labels](https://github.com/Ravenbrook/mps/labels) to the importance document.

Issue #205, opened 2023-03-29
The shield is not checked in cool builds


rptb1:
Change d82da0e1401d5541a9fbba78ca2bd5b03cb0d6c1 refactored the shield into its own structure, ShieldStruct, which makes sense in terms of abstractions.  [Various checks that were part of GlobalCheck](https://github.com/Ravenbrook/mps/commit/d82da0e1401d5541a9fbba78ca2bd5b03cb0d6c1#diff-8a34b9342ebc6fbd4d47e3d94dcc3e65871b738830a2a56f257a2c673bb213abL157-L189) were [moved to the new ShieldCheck](https://github.com/Ravenbrook/mps/commit/d82da0e1401d5541a9fbba78ca2bd5b03cb0d6c1#diff-32e19141eb9cbf244e80f8ad6859333c93af022817ccd8b252aad8c7e2cf8922R45-R83) and shield checking was then done [in the usual way by a CHECKD](https://github.com/Ravenbrook/mps/commit/d82da0e1401d5541a9fbba78ca2bd5b03cb0d6c1#diff-8a34b9342ebc6fbd4d47e3d94dcc3e65871b738830a2a56f257a2c673bb213abR157) .  The trouble with this change is that ShieldCheck is never reached in normal builds.

There are several problems:
1. CHECKD does not recurse in a cool (debugging) build, because by default that build uses CheckLevelSHALLOW.  This means that ArenaCheck does not invoke ShieldCheck.
2. Important shield functions such as ShieldEnter take an arena instead of a shield, then check that.  But that doesn't invoke ShieldCheck (see above).
3. We do not regularly run tests with CheckLevelDEEP which would cause CHECKD to recurse.

With several possible solutions:
1. CHECKD might be the wrong thing for a single-instance inlined structure such as ShieldStruct within ArenaStruct.
2. Shield functions should take a shield and check it.
3. We should build and run a deep checking build on at least one platform.

On this last point, a build with ``make -f lii6ll.gmk VARIETY=cool CFLAGSDEBUG='-O0 -g3 -DCHECKLEVEL=CheckLevelDEEP' testci`` [currently](https://github.com/Ravenbrook/mps/commit/179341b6cc2d9ab623ea53d10914d66c3668d47d) fails in several tests.

This issue was discovered during a walkthrough of https://github.com/Ravenbrook/mps/commit/e47bc4c04 with @rptb1 and @thejayps .

[...]

PR #207, opened 2023-03-29
Improvements to shield checking

### LINKED ISSUE #206: Tests assert with SHIELD_DEBUG set
`make -f lii6ll.gmk VARIETY=cool CFLAGSDEBUG='-O0 -ggdb3 -DSHIELD_DEBUG' testci` at https://github.com/Ravenbrook/mps/commit/179341b6cc2d9ab623ea53d10914d66c3668d47d produces multiple assertion failures in various tests, all of the form:
```
shield.c:644: MPS ASSERTION FAILED: queued == shield->limit
```

It is likely that this assertion is incorrect, but this needs thorough investigation.  Errors in shield invariants are very costly.

Found while checking the validity of https://github.com/Ravenbrook/mps/commit/e47bc4c04 .

We should also run a build with SHIELD_DEBUG set as part of CI, to prevent similar problems early.

rptb1:
Working on #205 .

Fixes #206 .

### DIFF
--- a/code/seg.c
+++ b/code/seg.c
@@ -185,8 +185,6 @@ static void segAbsFinish(Inst inst)
 
   AVERT(Seg, seg);
 
-  RingRemove(SegPoolRing(seg));
-
   arena = PoolArena(SegPool(seg));
 
   /* TODO: It would be good to avoid deprotecting segments eagerly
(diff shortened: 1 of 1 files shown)


rptb1:
Merged into #209 .

Issue #208, opened 2023-03-30
Testing with deep checking reports false positives


rptb1:
#205 says:

> ... a build with make -f lii6ll.gmk VARIETY=cool CFLAGSDEBUG='-O0 -g3 -DCHECKLEVEL=CheckLevelDEEP' testci [currently](https://github.com/Ravenbrook/mps/commit/179341b6cc2d9ab623ea53d10914d66c3668d47d) fails in several tests.

and also:

> We should build and run a deep checking build on at least one platform.

Several steps are needed:

1. Introduce an easier way to build and test with deep checking.
2. Add testing with deep checking to CI so that we catch problems early.
3. Analyse and fix current test failures with deep checking.

I have found false positives in deep checking errors have been introduced in various changes:

1. [branch/2016-04-16/trace-gens](https://github.com/Ravenbrook/mps/tree/branch/2016-04-16/trace-gens) [finishes the ring of generations](https://github.com/Ravenbrook/mps/blob/branch/2016-04-16/trace-gens/code/trace.c#L788) but goes on to [use the arena with an invalid trace](https://github.com/Ravenbrook/mps/blob/branch/2016-04-16/trace-gens/code/trace.c#L790-L800).
2. [branch/2020-08-31/walk](https://github.com/Ravenbrook/mps/tree/branch/2020-08-31/walk) [synthesizes a trace](https://github.com/Ravenbrook/mps/blob/branch/2020-08-31/walk/code/walk.c#L437-L440) that fails to pass [TraceCheck](https://github.com/Ravenbrook/mps/blob/branch/2020-08-31/walk/code/trace.c#L200) because it has no generations.
3. Deep checking has an infinite recursion in ArenaCheck via ShieldCheck, which checks some segments, which check their addresses via ChunkOfAddr, which checks the arena with ArenaCheck.  I haven't discovered when this was introduced yet.
4. ArenaRootsWalk synthesizes an invalid trace in TraceStateINIT that is flipped.  ArenaRootsWalk dates back to 1999 so it was introduced by some other change.  I haven't discovered when yet.

There may be more.  I have yet to run the entire test suite.

So far I have no evidence that any of these would cause failures in production, but deep checking is a valuable tool when, e.g. working on subtle parts of the MPS like the shield, and we ought to fix it up.

rptb1:
[...]

PR #209, opened 2023-03-30
Fixing deep checking

rptb1:
Working on #208 .

### DIFF
--- a/code/mpmst.h
+++ b/code/mpmst.h
@@ -669,13 +669,14 @@ typedef struct ShieldStruct {
   Sig sig;           /* design.mps.sig.field */
   BOOLFIELD(inside); /* <design/shield#.def.inside> */
   BOOLFIELD(suspended); /* mutator suspended? */
-  BOOLFIELD(queuePending); /* queue insertion pending? */
   Seg *queue;        /* queue of unsynced segs */
   Count length;      /* number of elements in shield queue */
   Index next;        /* next free element in shield queue */
   Index limit;       /* high water mark for cache usage */
   Count depth;       /* sum of depths of all segs */
   Count unsynced;    /* number of unsynced segments */
+  Count unqueued;    /* number of unsynced unqueued segments */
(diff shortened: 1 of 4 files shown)


rptb1:
#208 says:

> 1. [branch/2016-04-16/trace-gens](https://github.com/Ravenbrook/mps/tree/branch/2016-04-16/trace-gens) [finishes the ring of generations](https://github.com/Ravenbrook/mps/blob/branch/2016-04-16/trace-gens/code/trace.c#L788) but goes on to [use the arena with an invalid trace](https://github.com/Ravenbrook/mps/blob/branch/2016-04-16/trace-gens/code/trace.c#L790-L800).

Fixed in https://github.com/Ravenbrook/mps/pull/209/commits/01422138ed9699f93147b7b0927cb7f063f95a1b

rptb1:
#208 says:

> 3\. Deep checking has an infinite recursion in ArenaCheck via ShieldCheck, which checks some segments, which check their addresses via ChunkOfAddr, which checks the arena with ArenaCheck.  I haven't discovered when this was introduced yet.

Fixed in https://github.com/Ravenbrook/mps/pull/209/commits/54b93517c0f256366234c60e2d3bcc4cdd150686

rptb1:
Changelist 191812 does not appear to fix [job004026](https://www.ravenbrook.com/project/mps/issue/?action=job&job=job004026), as this assertion fires in many tests https://github.com/Ravenbrook/mps/blob/179341b6cc2d9ab623ea53d10914d66c3668d47d/code/shield.c#L88-L90

The comment on that assertion is false.  I think it *used* to be true, but various tweaks to the shield have rendered it ineffective.

Fixed in https://github.com/Ravenbrook/mps/pull/209/commits/0d56581f0d64d9f925fdc12c3288306d5dfbf689 by keeping counts of the number of synced segments in the queue, and the number of unsynced segments not in the queue, allowing exact checking of the queue length.

Issue #98, opened 2023-01-08
The MPS is hosted in Ravenbrook's Perforce repo, making it hard to collaborate


rptb1:
The true master copy of the MPS and all its history is currently in Ravenbrook's Perforce SCM repository, hosted on Ravenbrook's servers, and only accessible to Ravenbrook staff.  The Git repositories of the MPS (such as [the one at GitHub](https://github.com/Ravenbrook/mps)) are mirrors maintained by Ravenbrook using a combination of [Perforce Git Fusion](https://www.perforce.com/manuals/git-fusion/) and [a bit of automation](https://info.ravenbrook.com/infosys/robots/gitpushbot/).

Perforce is an excellent system, but the open source world now expects to collaborate using Git.  And in any case, Perforce requires per-seat licensing that we can't extend to the world.

Git Fusion is an amazing piece of work that allows for two-way synchronisation between Perforce and Git, but it is no longer maintained by Perforce, and it was never designed to maintain a permanent two-way mirror.  It still introduces friction in collaborating in Git, and has maintenance costs.

So, we plan to migrate the MPS to Git.  Initially, we will also be using GitHub as the hosting platform.

The first step will involve switching the "home" of the MPS to Git.  This means treating the Git repo as the "source of truth" and Perforce as the mirror.  Work will primarily take place using Git (and via GitHub).  We have to update our procedures and infrastructure to support Git.  And we should, as far as possible, make all of it accessible and usable by anyone, not just Ravenbrook staff.  Then we will truly have opened up the MPS to FOSS collaboration.

Git Fusion has created a complete mirror of the history of the MPS codeline, including [changes going back to 1994](https://github.com/Ravenbrook/mps/commit/16f1343b84a912e34c5a34023136141b18a2d516).  There is other material that hasn't made it into the Git repo (see #95 for some of it) but that can also be migrated.  There is other very important material, such as the issues data (stored in Perforce jobs), that has no home in Git, but might be mirrored to GitHub, but we must be very cautious about that dependency.

In the short term, we are likely to keep using Git Fusion to keep Perforce up-to-date.  It may be possible to do this permanently with little overhead, but it should not be *necessary* to the future of the MPS.  But it's likely that the MPS in Perforce will suffer a divorce with Git, and be mothballed.

This issue exists for discussing and tracking the migration.

rptb1:
[...]

PR #228, opened 2023-05-30
Updating the merge procedure to removing synchronisation with Perforce repository

rptb1:
Progress towards #98 .

### DIFF
--- a/procedure/pull-request-merge.rst
+++ b/procedure/pull-request-merge.rst
@@ -52,8 +52,8 @@ The purpose of this procedure is:
 
 2. protect the MPS from introduction of defects
 
-3. maintain and protect the permanent record of the MPS in Perforce
-   and Git, so that defects can be discovered, fixed, and prevented
+3. maintain and protect the permanent record of the MPS in Git, so
+   that defects can be discovered, fixed, and prevented
 
 As with any procedure, you can vary this one to meet this purpose, but
 you should probably read section "`6. Rationale`_".
@@ -101,11 +101,7 @@ When you finish the checklist, decide whether to start
 
 #. Is the merge approved by an approving review?
 
-   The `Ravenbrook MPS repo on GitHub`_ is set to require an approving
-   review and GitHub will block a push (step 8) without one.  If you
-   push an unapproved merge to Perforce, this will cause gitpushbot to
-   fail, and leave Perforce and GitHub out of sync.  *Do not push to
-   Perforce without GitHub approval.*
+   [FIXME: Should this refer to proc.review.exit?  RB 2023-05-30]
 
 #. Has the contributor licensed their work?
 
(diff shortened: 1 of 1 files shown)


thejayps:
Begin checking at 1646, resume at 1700 with 10 mins checking and small break

UNAA008:
m1 "Ravenbrook is currently (2023-01) "
   The edits have simplified the procedure already.
   Delete this paragraph, or mention the migration for historical note.

q1 Will hyperlinks to the merging procedure still work properly after
merging?

thejayps:
Stop Express review: major defect M1:

In section 6:
" git pull --rebase perforce
then go back to testing (step 4).

Alternatively, you could undo your merging work:

git reset --hard perforce/master "

These commands make reference to a token (remote name) "perforce". The purpose of this change is explicitly to remove synchronisation with perforce. Anything that still refers to "perforce" should either explain why the reference must still exist, or change the reference. If not done, this is sufficiently unclear that even a very experienced git/perforce user who could figure out what was going on would probably abort the procedure to be safe.

thejayps:
Stopped at 1710

rptb1:
> Stop Express review: major defect M1:
> 
> In section 6: " git pull --rebase perforce then go back to testing (step 4).
> 
> Alternatively, you could undo your merging work:
> 
> git reset --hard perforce/master "
> 
> These commands make reference to a token (remote name) "perforce". The purpose of this change is explicitly to remove synchronisation with perforce. Anything that still refers to "perforce" should either explain why the reference must still exist, or change the reference. If not done, this is sufficiently unclear that even a very experienced git/perforce user who could figure out what was going on would probably abort the procedure to be safe.

I think s/perforce/github/ fixes this but I'll want to validate those cases given Git's constant v violation of [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

rptb1:
> ...validate those cases given Git's constant violation of [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Yep, Git astonished again.  `git pull --rebase` discards merge commits, effectively fast-forwarding master.  You have to say `git pull --rebase=merges`.

thejayps:
Attempt 2 at express review: 

Executing [proc.review.entry](https://github.com/Ravenbrook/mps/blob/0ee19fffed94be25fe409f07f6ac9f61a007de50/procedure/review.rst)

Start time 1350

entry.express: still looks low risk, await @UNAA008 

1400 @UNAA008 arrives

Most comments relating to entry and planning here https://github.com/Ravenbrook/mps/pull/228#issuecomment-1779568832 remain the same, however the restructured text syntax check and FIXME checks now fail for the most recent commit to #123, so we are aware that there is some degradation in the source document for the review process. However we strongly suspect that this is cosmetic and/or superficial, and should not hinder an express review 

.express.major We *are* performing an express review again after a major defect was found. The author @rptb1 has made edits to fix the major defect found, and we believe express review is appropriate following this. Is the underlying procedure completely correct?

Lots of discussion points are coming up about the appropriateness of express review - on balance this is probably not for express review now.

We discussed an idea to test the merge procedure "in anger" on a small low risk change, then do a full or express review after that.

@UNAA008 points out that changes like this one, which document processes for maintaining the MPS, should be exercised as part of their inspection.

rptb1:
> We discussed an idea to test the merge procedure "in anger" on a small low risk change, then do a full or express review after that.

This procedure was exercised in 40b5eb72c799e73f94d5a1fef5ae0b439a7c054d for #283 .

> Yep, Git astonished again. `git pull --rebase` discards merge commits, changing the topology of what is rebased and effectively fast-forwarding master. You have to say `git pull --rebase=merges`.

This was anti-exercised in https://github.com/Ravenbrook/mps/pull/284#issuecomment-2503016258 for #284 resulting in 40b5eb72c799e73f94d5a1fef5ae0b439a7c054d..8d2019bb0db81709d162f712aefa68d55f622045 .

> Is the underlying procedure completely correct?

It's hard to tell, because Git.  However, I believe the risk to the MPS or its users is low, and an express review is appropriate.

PR #231, opened 2023-06-05
Automatically checking copyright dates in copyright notices

rptb1:
Experimental script to check the copyright dates in notices such as "Copyright (C) 1997-2023" and in particular that the end date is consistent with when the file was last modified in Git's history.  This could be incorporated into CI.


rptb1:
See also review issues in https://github.com/Ravenbrook/mps/pull/223#pullrequestreview-1432893665

thejayps:
Copyright dates can be very useful for archaeology where the history of a codebase is unclear

Issue #256, opened 2023-07-12
error: storing the address of local variable 'stackOut' in '*stackOut.0_1' [-Werror=dangling-pointer=]


topazus:
I try to compile mps 1.118.0 with GCC 13, but I occured the build error.

```
lii6gc: lii6gc/cool/ss.o
ss.c: In function 'StackHot':
ss.c:38:13: error: storing the address of local variable 'stackOut' in '*stackOut.0_1' [-Werror=dangling-pointer=]
   38 |   *stackOut = &stackOut;
      |   ~~~~~~~~~~^~~~~~~~~~~
ss.c:36:22: note: 'stackOut' declared here
   36 | void StackHot(void **stackOut)
      |               ~~~~~~~^~~~~~~~
ss.c:36:22: note: 'stackOut' declared here
cc1: all warnings being treated as errors
gmake[3]: *** [comm.gmk:628: lii6gc/cool/ss.o] Error 1
gmake[3]: Leaving directory '/home/ruby/rpmbuild/BUILD/mps-release-1.118.0/code'
gmake[2]: *** [comm.gmk:398: target] Error 2
gmake[1]: *** [comm.gmk:370: mps.a] Error 2
gmake[1]: *** Waiting for unfinished jobs....
```

thejayps:
Many thanks for your report. Can you tell us anything more about how what you're working on uses the MPS?

We currently build and test the MPS on Ubuntu 22, which is currently on GCC 11.3.0.  We'll investigate the GCC 13 build.

If you are not modifying the MPS itself, you can build it for your project by following 2.3. Compiling the MPS for your project [<https://memory-pool-system.readthedocs.io/en/version-1.118/guide/build.html#compiling-the-mps-for-your-project>](https://memory-pool-system.readthedocs.io/en/version-1.118/guide/build.html#compiling-the-mps-for-your-project).  This will not produce warnings as errors.

If you're developing the MPS, you can work around this problem locally by modifying code/gc.gmk to suppress this warning, by adding something like -Wno-error=dangling-pointer to CFLAGSCOMPILER.

If you are integrating the MPS with Ruby (or have done so), we'd very much like to hear about it.

waywardmonkeys:
In GitHub Actions, `ubuntu-latest` should have `gcc-12` / `g++-12` available and you can test multiple versions using a matrix. (It should have gcc 9, 10, 11, and 12 as well as clang 12, 13, 14.)

topazus:
[...]

PR #257, opened 2023-07-23
Resolve warnings when compiling with GCC 12

rptb1:


rptb1:
GCC 13.2 applied as shown in https://github.com/Ravenbrook/mps/issues/256#issuecomment-1712843937 still produces the error in #256 .

thejayps:
All checks pass but check list is out of date, relevant check is not included #258

rptb1:
We need to ensure the original problem is tested by CI.  See #258 or possibly we can use some other "latest" environment.

PR #258, opened 2023-07-23
Add Fedora CI

topazus:


thejayps:
performing proc.regular with @rptb1 

This is related Fedora Rawhide (https://docs.fedoraproject.org/en-US/releases/rawhide/). We need to work out what if anything should be gated on the results of this CI

thejayps:
Are there equivalent bleeding edge builds that we should consider? We need a policy for these in CI

PR #270, opened 2023-10-22
Experimenting with rendering the manual using rinohtype

rptb1:
The [rinohtype](https://github.com/brechtm/rinohtype/tree/0550a0d202cb498c1aac3bac1f6920e7fbbf3c27) package is mentioned in the [Sphinx FAQ](https://www.sphinx-doc.org/en/master/faq.html#how-do-i) as a way of rendering docs into PDF without using LaTeX.  This is just a quick attempt at doing that.  It doesn't work (yet).  See https://github.com/brechtm/rinohtype/issues/421 .


thejayps:
We don't really need a pdf manual. It would be nice to have

PR #283, opened 2024-02-29
Update config.guess to the version from autoconf 2.72

waywardmonkeys:
Fixes #281. Identification of newer systems such as macOS on Apple Silicon was not working correctly with the old version of this file.

Issue #281, opened 2024-02-29
Configuring on M1 mac fails (macOS 14)


gerd-moellmann:
Configuring mps (`./configure ...`) on an M1 mac with macOS 14.3.1 fails for me. Making the change below and regenerating `configure` with `autoconf` lets me build, and all tests succeeded.

```
modified   configure.ac
@@ -91,6 +91,17 @@ case $host/$CLANG in
       TEST_TARGET=test-xcode-build
       PFMCFLAGS="$CFLAGS_LL"
     ;;
+  arm-apple-darwin*/*)
+      AC_MSG_RESULT([Mac OS X ARM64])
+      MPS_OS_NAME=xc
+      MPS_ARCH_NAME=a6
+      MPS_BUILD_NAME=ll
+      BUILD_TARGET=build-via-xcode
+      CLEAN_TARGET=clean-xcode-build
+      INSTALL_TARGET=install-xcode-build
+      TEST_TARGET=test-xcode-build
+      PFMCFLAGS="$CFLAGS_LL"
+    ;;
   i*86-*-darwin*/*)
       AC_MSG_RESULT([Mac OS X x86])
       MPS_OS_NAME=xc

```

waywardmonkeys:
It is odd, but I get `arm` there as well. (Not `arm64`, not `aarch64`.)

Support for this platform was previously added in e5066a7c7a76103875d89c55c4770b3d989bf61e ... so I'm curious if this changed since then or what.

waywardmonkeys:
I've done some analysis and will submit a PR momentarily for what I found.

waywardmonkeys:
See PR #283.

gerd-moellmann:
Bruce Mitchener ***@***.***> writes:

> See PR #283.

Thanks!

PR #289, opened 2024-07-24
Update to autotools 2.72

skangas:
This provides a more complete fix for #281 than PR #283 does. I've also included the manual steps below, which should be easily reproducible. This is an alternative to merging this patch outright (depending on your level of cautiousness after the xz backdoor).

Please see below the commit message.

Thanks!

---

This essentially just does the following, with autotools 2.72 installed:

    $ cp /usr/lib/share/autoconf/build-aux/config.guess \
        tool/autoconf/build-aux/config.guess
    $ cp /usr/lib/share/autoconf/build-aux/config.sub \
        tool/autoconf/build-aux/config.sub
    $ autoreconf
    configure.ac:28: warning: The macro 'AC_LANG_C' is obsolete.
    configure.ac:28: You should run autoupdate.
    ./lib/autoconf/c.m4:72: AC_LANG_C is expanded from...
    configure.ac:28: the top level
    $ autoupdate

After running autoupdate, the following manual fix was needed:

    -        [m4_esyscmd_s(sed -n '/^#define MPS_RELEASE "\(.*\)"$/{s/.*"\(.*\)"/\1/;p;}' code/version.c)],
    +        [m4_esyscmd_s([sed -n '/^#define MPS_RELEASE "\(.*\)"$/{s/.*"\(.*\)"/\1/;p;}' code/version.c])],


rptb1:
1. Checking started 13:37.
2. Checking changes against [Autoconf documentation](https://web.archive.org/web/20240925080748/http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/autoconf.html).
3. AC_LANG_C is indeed marked as obsolete in the macro index and removed in https://github.com/Ravenbrook/mps/pull/289/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810L28-R28 . AC_LANG is documented [here](https://web.archive.org/web/20240925080748/http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/autoconf.html#index-AC_005fLANG).
4. The [Wayback Machine documentation](https://web.archive.org/web/20240925080748/http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/autoconf.html#index-AC_005fLANG) is for Autoconf 2.72 which matches [what's specified in the change](https://web.archive.org/web/20240925080748/http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/autoconf.html#index-AC_005fLANG).
5. Checked out the branch and ran configure locally on Kiwi-Ubuntu.
6. Autoconf on Ubuntu 24 is version 2.71, so `autoreconf` cannot run there.  This means `autoreconf` cannot be run with this change on a supported platform.
7. Halting review with a major defect.

rptb1:
There are possibly some undocumented requirements and step here that have led to this change failing review.  We'll back to that.  Essentially, we need to decide whether we require autoreconf to work on target platforms, of it's acceptable for there to be extra steps (such as installing a later GNU autoconf).

waywardmonkeys:
That might be why I only did an update of `config.guess` as my ancient memories are that `config.guess` can be updated separately from the rest and that used to be a recommendation some 20-30 years ago when porting to newer systems.

rptb1:
> That might be why I only did an update of `config.guess` as my ancient memories are that `config.guess` can be updated separately from the rest and that used to be a recommendation some 20-30 years ago when porting to newer systems.

Your recollection is supported by https://web.archive.org/web/20240813003154/https://www.gnu.org/software/gettext/manual/html_node/config_002eguess.html which says:

> You can obtain the newest version of config.guess and config.sub from the ‘config’ project at https://savannah.gnu.org/.

rptb1:
The MPS does not really need autoconf.  About the only thing we use it for is to get the target platform, which is the output of config.guess, just to offer an easy "./configure && make" command line.  We could possibly go with our own, very simple, configure script and remove the autoconf dependency.  While autoconf is around there's a risk that someone might start using it!

PR #293, opened 2024-11-26
Updating checkout actions to one permitted by GitHub

rptb1:
GitHub builds started failing on MPS branches because they referenced a checkout action that has been [forbidden by GitHub for security](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/).

The scope of this branch is likely to expand to cover other problems with the automated GitHub builds, which appear to have rotted somewhat.


waywardmonkeys:
I had updated a file in addition to what you did here

